PR #3, opened 2023-03-20
Modular Forge API & Patching

BluSpring:
Separates the Forge API into its own source set, and adds code patching to make sure that updating Kilt to newer Forge versions is as painless as possible.


BluSpring:
I just realized I've fully accomplished this entirely ages ago, lemme just. merge this in.

PR #18, opened 2024-08-07
Fix model loading and implement sodium compat

AlphaMode:


BluSpring:
Merged manually, don't worry

PR #107, opened 2024-12-23
Support Gradle configuration cache, small buildscript cleanups

dynpatch:
Hope this isn't too opinionated, feel free to ask for changes!

PR #112, opened 2024-12-23
New ModLauncher implementation, moved into its own submodule

dynpatch:
Reimplemented the ModLauncher with classes taken directly from source, brought back removed API interfaces between 10.x and 11.x as default methods for better compat, `IEnvironment` is now filled with every key (except `AUDITTRAIL` since it'd be a bit more complex and very few mods use it).

This PR also moves the implementation to its standalone Gradle submodule, as it doesn't interact with the rest of the project and is there for compat only. I'd love some feedback on this, as I feel code splitting and separation of concerns are key in a project of this scale.

Note that for now the buildscript part of the """modular setup""" is very rudimentary and is just there to kick things into action for progressive modular adoption (i.e small/kilt-related parts, like loader & apis, not Forge as a whole like Patchwork did).


kilt-modlauncher/build.gradle.kts
```diff
@@ -0,0 +1,10 @@
+repositories {
+    mavenCentral()
+    maven("https://maven.neoforged.net/releases")
+    maven("https://maven.fabricmc.net/")
+}
+
+dependencies {
+    compileOnly("cpw.mods:securejarhandler:3.0.4")
+    compileOnly("net.fabricmc:fabric-loader:0.14.9")
```
BluSpring: These should use the versions provided in Kilt's gradle.properties.

settings.gradle.kts
```diff
@@ -1,3 +1,5 @@
+enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
```
BluSpring: What's this for?

settings.gradle.kts
```diff
@@ -16,3 +18,7 @@ pluginManagement {
                     .split("+")[0] // Ensures sentences like `+build.1` are ignored
     }
 }
+
+rootDir.listFiles { f: File -> f.isDirectory && f.name.startsWith("kilt-") }?.forEach { dir ->
+    include(dir.name)
+}
```
BluSpring: I would prefer if this was a manual detection rather than an automatic detection.
Or, otherwise, place them into a "modules" folder for better organization.

src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoader.kt
```diff
@@ -381,7 +383,7 @@ class KiltLoader {
                 if (FabricLoader.getInstance()
                         .isModLoaded(mod.modId) || FabricLoaderImpl.INSTANCE.getModCandidate(mod.modId) != null
                 ) {
-                    Kilt.logger.warn("Duplicate Forge and Fabric mod IDs detected: ${mod.modId}")
+                    Kilt.logger.warn("Duplicate Fabric and Forge mod IDs detected: ${mod.modId}")
```
BluSpring: I don't see any point in switching these around, honestly.

build.gradle.kts
```diff
@@ -129,6 +134,9 @@ dependencies {
         exclude("net.fabricmc", "fabric-loader")
     })
 
+    // Kilt modules
+    include(implementation(projects.kiltModlauncher)!!)
```
BluSpring: If this is going to be included, it might be worth making an FMJ for the module so ModMenu can detect it and provide the correct license, name and credits.

kilt-modlauncher/build.gradle.kts
```diff
@@ -0,0 +1,10 @@
+repositories {
+    mavenCentral()
+    maven("https://maven.neoforged.net/releases")
+    maven("https://maven.fabricmc.net/")
+}
```
BluSpring: Aren't these already defined in the buildscript?

kilt-modlauncher/README.md
```diff
@@ -0,0 +1,7 @@
+# kilt-stub-modlauncher
+
+A reimplementation of the [modlauncher](https://github.com/McModLauncher/modlauncher) API, SPI, and common classes for Kilt.
+
+## Licensing
+
+This module includes parts of [modlauncher](https://github.com/McModLauncher/modlauncher/), which is licensed under the [GPLv3](https://github.com/McModLauncher/modlauncher/blob/10.0/COPYING).
```
BluSpring: ModLauncher is licensed under the [LGPLv3](https://github.com/McModLauncher/modlauncher/blob/main/LICENSE), not the GPLv3. That was, however, what it was licensed as previously, I believe, or there are some components that are under GPLv3, not sure.

BluSpring (changes requested):
Left some of my thoughts on everything.

kilt-modlauncher/build.gradle.kts
qtdyn: I thought I fixed those, didn't push my bad

kilt-modlauncher/README.md
qtdyn: That's the license of the branch I used from (see the license link), I don't think the LGPL switch retroactively applies so I stuck with that for now.

kilt-modlauncher/build.gradle.kts
qtdyn: With #107 this is redundant yes

build.gradle.kts
qtdyn: Sure, I was just planning to do this simply as this isn't a full module setup, but i'd be down to do more changes to make this work

src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoader.kt
qtdyn: That was just an IDE warning for duplicate log messages, albeit not that useful since this log probably won't get reached if the earlier (duplicate) one is called (i think?)

settings.gradle.kts
qtdyn: Which would you prefer?

settings.gradle.kts
qtdyn: This is to enable using project accessors like [here](https://github.com/KiltMC/Kilt/pull/112/files#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06R138)

kilt-modlauncher/README.md
BluSpring: The 10.0 branch is also using LGPL - https://github.com/McModLauncher/modlauncher/blob/10.0/LICENSE
This is also displayed by the fact that all the files that have been included into this PR are in fact licensed under LGPL.

kilt-modlauncher/build.gradle.kts
BluSpring: Might be worth removing those then now that #107 has been merged.

src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoader.kt
BluSpring: Unfortunately this one is required, as sometimes it manages to miss that first one, hence why the second one is included.

settings.gradle.kts
BluSpring: The "modules" folder would definitely be better, since that handles automatically adding them without needing to filter by a specific prefix.

In which case, I think `kilt-modlauncher` would work better as `modules/modlauncher`.

kilt-modlauncher/README.md
qtdyn: im dumb my bad

.github/workflows/gradle.yml
```diff
@@ -17,7 +17,6 @@ jobs:
     - uses: actions/checkout@v4
       with:
         submodules: true
-        token: ${{ secrets.PAT_TOKEN }}
```
BluSpring: This is required - otherwise, GH Actions seems to be unable to get the submodules, from previous experience with this repo.
Not entirely sure why, but it's needed.

BluSpring (changes requested):

buildSrc/src/main/kotlin/xyz/bluspring/kilt/gradle/util/dsl.kt
BluSpring, line 1: I would recommend this file follow the naming conventions of all the other Kotlin files - so use the name "DSL.kt" or "DSLUtils.kt" rather than "dsl.kt".

src/main/resources/kilt.accesswidener
```diff
@@ -1,13 +1,4 @@
 accessWidener v2 named
-
-# Kilt's own access wideners.
-accessible class net/minecraft/server/network/ServerLoginPacketListenerImpl$State
-
-accessible method net/minecraft/client/renderer/texture/SpriteContents$AnimatedTexture getFrameX (I)I
-accessible method net/minecraft/client/renderer/texture/SpriteContents$AnimatedTexture getFrameY (I)I
-transitive-accessible method net/minecraft/world/entity/Entity setRemoved (Lnet/minecraft/world/entity/Entity$RemovalReason;)V
-transitive-extendable method net/minecraft/world/entity/Entity setRemoved (Lnet/minecraft/world/entity/Entity$RemovalReason;)V
-
```
BluSpring, line 10: Should probably have these automatically added by the TransformerToWidener task, as these are vital AWs.

buildSrc/src/main/kotlin/xyz/bluspring/kilt/gradle/task/TransformerToWidernerTask.kt
BluSpring, line 1: The file is misspelled - should be TransformerToWidenerTask.kt.

buildSrc/src/main/kotlin/xyz/bluspring/kilt/gradle/java.gradle.kts
```diff
@@ -0,0 +1,81 @@
...
+//            artifact(kotlinSourcesJar) {
+//                builtBy(remapSourcesJar)
+//            }
+//        }
+//    }
+//
+//    repositories {
+//        mavenLocal()
+//    }
+//}
```
BluSpring, line 81: Try to leave publishing enabled - in the future, I plan to have Kilt as an available helper towards better Forge mod support within Fabric mods.

modules/README.md
```diff
@@ -0,0 +1,3 @@
+# kilt modules
```
BluSpring, line 1: Use propercasing - "Kilt Modules".

BluSpring (changes requested):

buildSrc/src/main/kotlin/xyz/bluspring/kilt/gradle/java.gradle.kts
qtdyn: This was WIP, will do :+1:

src/main/resources/kilt.accesswidener
qtdyn: That probably wasn't intentional, will look into it

buildSrc/src/main/kotlin/xyz/bluspring/kilt/gradle/task/TransformerToWidernerTask.kt
qtdyn: typing hard :pensive:

PR #206, opened 2025-04-04
Make Java version become Java provider

Cartrigger:
We should really be looking at the provider vs Java version in these reports, as we can easily check what version of Java the person is using as it gets logged along with mods. If the person isn't providing something that is listing their mods, theres something wrong with their report. There shouldn't be too many differences between providers like OpenJDK, Oracle, Azul, and Temurin, but there are differences, and some have more features unlocked for certain versions than others, or may optimize differently, like Oracle's GraalVM JDK version.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -74,8 +74,8 @@ body:
   - type: input
     id: java
     attributes:
-      label: Java version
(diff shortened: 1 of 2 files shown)


BluSpring:
Maybe include the version as well? As in "Java Provider + Version", with the example "Eclipse Temurin 17.0.6"

Cartrigger:
yeah ill update it to do be that tbh

PR #486, opened 2025-10-04
Fire the Forge BreakEvent.

Acuadragon100:
This patch makes Kilt fire the Forge BreakEvent. This event is fired before a player tries to break a block, giving mods accepting the event the option to cancel the event (this leaving the block intact) or set an arbitrary amount of xp to drop if the block was successfully broken.

This patch is important for [Stargate Journey](https://modrinth.com/mod/sgjourney), as otherwise it is effectively impossible to remove frame cover blocks from the stargate.

For some reason, the game seems to ignore the return value of the WrapOperation mixin around canAttackBlock whenever the player's main hand is empty. I don't know why this happens, but I'm guessing it's related to some optimisation happening behind the scenes (branch prediction maybe?). This is why I added the CallbackInfoReturnable.

For the second injection I unfortunately had to use a shift statement because it simply could not find the local variables from the TAIL position otherwise.

For now this mixin uses the default priority (999). Something I was considering was to set the priority parameter even lower since any other mods wrapping canAttackBlock with a lower value of the priority parameter would be silently ignored. This could also be resolved by writing a custom version of ForgeHooks::onBlockBreakEvent which takes the callback as an argument, but then this would have to be updated manually whenever it the original Forge method is changed.

### DIFF
--- a/src/main/java/xyz/bluspring/kilt/forgeinjects/server/level/ServerPlayerGameModeInject.java
+++ b/src/main/java/xyz/bluspring/kilt/forgeinjects/server/level/ServerPlayerGameModeInject.java
@@ -7,6 +7,7 @@ import com.llamalad7.mixinextras.sugar.Cancellable;
 import com.llamalad7.mixinextras.sugar.Local;
 import com.llamalad7.mixinextras.sugar.Share;
 import com.llamalad7.mixinextras.sugar.ref.LocalIntRef;
+import net.fabricmc.loader.api.FabricLoader;
 import net.minecraft.core.BlockPos;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.server.level.ServerPlayer;
@@ -42,17 +43,19 @@ public abstract class ServerPlayerGameModeInject {
 			at = @At(
 					value = "INVOKE",
 					target = "Lnet/minecraft/world/item/Item;canAttackBlock(Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/entity/player/Player;)Z"
-			) // We want to use WrapOperation or Redirect to ensure canAttackBlock is not fired more than once (it might have side effects).
+			)
 	)
-	public boolean canMine(
+	private boolean kilt$canAttackBlock(
 			Item instance, BlockState state, Level level, BlockPos pos, Player player, Operation<Boolean> original,
 			@Cancellable CallbackInfoReturnable<Boolean> cir, @Share("exp") LocalIntRef exp
 	) {
-		int expLocal = ForgeHooks.onBlockBreakEvent(level, gameModeForPlayer, this.player, pos);
+		int expLocal = ForgeHooks.kilt$onBlockBreakEvent(
+				level, gameModeForPlayer, this.player, pos,
+				() -> original.call(instance, state, level, pos, player)
+		);
 		if (expLocal == -1) {
-			// When the player's main hand item is empty, the game seems to ignore the value returned by the mixin and breaks the block anyway.
-			// Might only affect certain hardware.
-			if (player.getMainHandItem().isEmpty()) {
+			// Patch for porting lib still breaking block when event is cancelled with an empty main hand item.
+			if (FabricLoader.getInstance().isModLoaded("porting_lib_base") && player.getMainHandItem().isEmpty()) {
 				cir.setReturnValue(false);
 			}
 			return false;
@@ -68,7 +71,7 @@ public abstract class ServerPlayerGameModeInject {
 					shift = At.Shift.BEFORE // Necessary to capture variables.
 			)
 	)
-	public void dropXP(
+	private void kilt$dropXP(
 			BlockPos pos, CallbackInfoReturnable<Boolean> cir,
 			@Local BlockState blockState,
 			@Local(ordinal = 0) boolean removed, @Share("exp") LocalIntRef exp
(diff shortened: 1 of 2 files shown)


src/main/java/xyz/bluspring/kilt/forgeinjects/server/level/ServerPlayerGameModeInject.java
```diff
@@ -1,10 +1,81 @@
...
+		if (expLocal == -1) {
+			// When the player's main hand item is empty, the game seems to ignore the value returned by the mixin and breaks the block anyway.
+			// Might only affect certain hardware.
+			if (player.getMainHandItem().isEmpty()) {
+				cir.setReturnValue(false);
+			}
+			return false;
+		}
+		exp.set(expLocal);
+		return true;
```
BluSpring: The `original` operation isn't called here, which results in this mixin not supporting any mods that would `WrapOperation` here themselves.

I recommend adding a parameter into `onBlockBreakEvent` (and thus making a new method `kilt$onBlockBreakEvent`, reference other Kilt methods in `ForgeHooks` / `ForgeHooksClient`) to solve this problem.

src/main/java/xyz/bluspring/kilt/forgeinjects/server/level/ServerPlayerGameModeInject.java
```diff
@@ -1,10 +1,81 @@
...
+	protected ServerLevel level;
+
+	@WrapOperation(
+			method = "destroyBlock",
+			at = @At(
+					value = "INVOKE",
+					target = "Lnet/minecraft/world/item/Item;canAttackBlock(Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/entity/player/Player;)Z"
+			) // We want to use WrapOperation or Redirect to ensure canAttackBlock is not fired more than once (it might have side effects).
+	)
+	public boolean canMine(
```
BluSpring: Make method private, and add `kilt$` prefix to the method.

src/main/java/xyz/bluspring/kilt/forgeinjects/server/level/ServerPlayerGameModeInject.java
```diff
@@ -1,10 +1,81 @@
...
+	}
+
+	@Inject(
+			method = "destroyBlock",
+			at = @At(
+					value = "TAIL",
+					shift = At.Shift.BEFORE // Necessary to capture variables.
+			)
+	)
+	public void dropXP(
```
BluSpring: Make method private, and add `kilt$` prefix to the method.

src/main/java/xyz/bluspring/kilt/forgeinjects/server/level/ServerPlayerGameModeInject.java
```diff
@@ -1,10 +1,81 @@
...
+			Item instance, BlockState state, Level level, BlockPos pos, Player player, Operation<Boolean> original,
+			@Cancellable CallbackInfoReturnable<Boolean> cir, @Share("exp") LocalIntRef exp
+	) {
+		int expLocal = ForgeHooks.onBlockBreakEvent(level, gameModeForPlayer, this.player, pos);
+		if (expLocal == -1) {
+			// When the player's main hand item is empty, the game seems to ignore the value returned by the mixin and breaks the block anyway.
+			// Might only affect certain hardware.
+			if (player.getMainHandItem().isEmpty()) {
+				cir.setReturnValue(false);
+			}
```
BluSpring: Check your mixin exports (`-Dmixin.debug.export=true`) and use breakpoints to see if this is absolutely necessary.

src/main/java/xyz/bluspring/kilt/forgeinjects/server/level/ServerPlayerGameModeInject.java
```diff
@@ -1,10 +1,81 @@
...
+
+	@Shadow
+	protected ServerLevel level;
+
+	@WrapOperation(
+			method = "destroyBlock",
+			at = @At(
+					value = "INVOKE",
+					target = "Lnet/minecraft/world/item/Item;canAttackBlock(Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/entity/player/Player;)Z"
+			) // We want to use WrapOperation or Redirect to ensure canAttackBlock is not fired more than once (it might have side effects).
```
BluSpring: `canAttackBlock` wouldn't be fired more than once, it's a boolean check, if a mod causes side effects then it's probably on them.

BluSpring (changes requested):
Kilt typically adds its own modifications into Forge classes anyway, you don't have to worry about that. The added methods are typically prefixed with `kilt$` so I wouldn't miss them in updating to newer versions.

You can make a second PR in Kilt's Forge fork to address this problem.

src/main/java/xyz/bluspring/kilt/forgeinjects/server/level/ServerPlayerGameModeInject.java
Acuadragon100: Ok, I looked into this a bit further. It looks like porting lib is the reason it doesn't work with empty item stacks (because porting lib has basically the same patch already, I just realised... It just isn't forwarded to the Forge event bus).

However, that means this issue could be solved by just adding a MinecraftForge.EVENT_BUS::post call to BlockEvents::BLOCK_BREAK in porting lib. Would that be a better approach?

BluSpring: In that case, I would also personally still resolve the review requests and have these injects commented out in the event that we have to disable/bypass Porting Lib.

BluSpring (approved):
I do have some tiny nitpick changes but I'll do them myself, I've been meaning to for a while now with Kilt.

To verify, you have ensured that everything continues to work correctly and any mods that require this change work with this PR, yes?

Acuadragon100:
I checked Stargate Journey and that works fine.
Also threw together a tiny test mod to make sure that xp drops so that should work too.
I didn't check any other mods though.

PR #526, opened 2025-10-23
Improve mod compatibility of re-equip check.

Acuadragon100:
This pull request makes the re-equip check slightly more inter-mod compatible.

You see, [Combatify](https://github.com/Alexandra-Myers/Combatify) and [CookeyMod](https://github.com/rizecookey/CookeyMod) have placed ModifyVariable mixins right in-between LocalPlayer::getAttackStrengthScale and Mth::clamp. When combined with Kilt, this causes an issue similar to Alexandra-Myers/Combatify#82 (important to note that the Kilt issue is different from 82 as Kilt was not present in the logs, but it shows similar symptoms).

I've double checked with a quickly thrown together test mod and it seems to continue working, but if you know any forge mods which has noticeable uses of this check then please let me know so I can test them as well.

### DIFF
--- a/src/main/java/xyz/bluspring/kilt/forgeinjects/client/renderer/ItemInHandRendererInject.java
+++ b/src/main/java/xyz/bluspring/kilt/forgeinjects/client/renderer/ItemInHandRendererInject.java
@@ -26,6 +26,7 @@ import net.minecraftforge.client.extensions.common.IClientItemExtensions;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.Unique;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
@@ -37,10 +38,6 @@ public abstract class ItemInHandRendererInject {
     @Shadow private ItemStack mainHandItem;
 
     @Shadow private ItemStack offHandItem;
-
-    @Shadow private float mainHandHeight;
-
-    @Shadow private float offHandHeight;
 
     @WrapWithCondition(method = "renderHandsWithItems", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/renderer/ItemInHandRenderer;renderArmWithItem(Lnet/minecraft/client/player/AbstractClientPlayer;FFLnet/minecraft/world/InteractionHand;FLnet/minecraft/world/item/ItemStack;FLcom/mojang/blaze3d/vertex/PoseStack;Lnet/minecraft/client/renderer/MultiBufferSource;I)V"))
     private boolean kilt$cancelRenderIfEventCancelled(ItemInHandRenderer instance, AbstractClientPlayer player, float partialTicks, float pitch, InteractionHand hand, float swingProgress, ItemStack stack, float equippedProgress, PoseStack poseStack, MultiBufferSource buffer, int combinedLight) {
@@ -64,12 +61,20 @@ public abstract class ItemInHandRendererInject {
             this.offHandItem = offhandStack;
     }
 
+	@Unique
+	private static boolean kilt$shouldSkipReequip(boolean fabricSkipReequip, boolean forgeSkipReequip) {
+		if (forgeSkipReequip) {
+			return true;
+		}
+		return fabricSkipReequip;
+	}
+
     @Definition(id = "mainHandItem", field = "Lnet/minecraft/client/renderer/ItemInHandRenderer;mainHandItem:Lnet/minecraft/world/item/ItemStack;")
     @Definition(id = "itemStack", local = @Local(type = ItemStack.class, ordinal = 0))
     @Expression("this.mainHandItem == itemStack")
     @ModifyExpressionValue(method = "tick", at = @At("MIXINEXTRAS:EXPRESSION"))
     private boolean kilt$useReequipCheckForMainHand(boolean original, @Share("reequipM") LocalBooleanRef reequipM) {
-		return !reequipM.get();
+		return kilt$shouldSkipReequip(original, !reequipM.get());
     }
 
 	@Definition(id = "offHandItem", field = "Lnet/minecraft/client/renderer/ItemInHandRenderer;offHandItem:Lnet/minecraft/world/item/ItemStack;")
@@ -77,7 +82,7 @@ public abstract class ItemInHandRendererInject {
 	@Expression("this.offHandItem == itemStack2")
 	@ModifyExpressionValue(method = "tick", at = @At("MIXINEXTRAS:EXPRESSION"))
 	private boolean kilt$useReequipCheckForOffHand(boolean original, @Share("reequipO") LocalBooleanRef reequipO) {
-		return !reequipO.get();
+		return kilt$shouldSkipReequip(original, !reequipO.get());
 	}
 
     @WrapOperation(method = "renderArmWithItem", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/item/ItemStack;is(Lnet/minecraft/world/item/Item;)Z", ordinal = 1))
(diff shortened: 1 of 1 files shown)


src/main/java/xyz/bluspring/kilt/forgeinjects/client/renderer/ItemInHandRendererInject.java
```diff
@@ -76,10 +77,13 @@ public abstract class ItemInHandRendererInject {
         return original.call(instance, item) || instance.getItem() instanceof CrossbowItem;
     }
 
-    @ModifyArg(method = "tick", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/Mth;clamp(FFF)F", ordinal = 2), index = 0)
-    private float kilt$useReequipCheckForMainHand(float original, @Share("reequipM") LocalBooleanRef reequipM, @Local LocalPlayer localPlayer) {
-        float f = localPlayer.getAttackStrengthScale(1.0F);
-        return (!reequipM.get() ? f * f * f : 0.0F) - this.mainHandHeight;
+    @ModifyExpressionValue(method = "tick", at = @At(value = "FIELD", target = "Lnet/minecraft/client/renderer/ItemInHandRenderer;mainHandItem:Lnet/minecraft/world/item/ItemStack;", ordinal = 2))
```
BluSpring: Actually, why not use the new `@Expression` system instead of simply targeting `this.mainHandItem` like that? You could safely target `this.mainHandItem == stack` and simply pass the boolean value.

I'm pretty sure this was written before Expressions came around, so it wasn't really written with great mod compatibility in mind.

src/main/java/xyz/bluspring/kilt/forgeinjects/client/renderer/ItemInHandRendererInject.java
```diff
@@ -76,10 +77,13 @@ public abstract class ItemInHandRendererInject {
...
+    private ItemStack kilt$useReequipCheckForMainHand(ItemStack original, @Local(ordinal = 0) ItemStack itemStack, @Share("reequipM") LocalBooleanRef reequipM) {
+        if (reequipM.get()) {
+            return null;
+        } else {
+            return itemStack;
+        }
     }
 
     @ModifyArg(method = "tick", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/Mth;clamp(FFF)F", ordinal = 3), index = 0)
```
BluSpring: Why not do the same thing for offhand too? Offhand seems to have the same mod compatibility problem.

BluSpring (changes requested):

src/main/java/xyz/bluspring/kilt/forgeinjects/client/renderer/ItemInHandRendererInject.java
```diff
@@ -62,28 +64,24 @@ public abstract class ItemInHandRendererInject {
...
+		return !reequipM.get();
+    }
 
-    }*/
+	@Definition(id = "offHandItem", field = "Lnet/minecraft/client/renderer/ItemInHandRenderer;offHandItem:Lnet/minecraft/world/item/ItemStack;")
+	@Definition(id = "itemStack2", local = @Local(type = ItemStack.class, ordinal = 1))
+	@Expression("this.offHandItem == itemStack2")
+	@ModifyExpressionValue(method = "tick", at = @At("MIXINEXTRAS:EXPRESSION"))
+	private boolean kilt$useReequipCheckForOffHand(boolean original, @Share("reequipO") LocalBooleanRef reequipO) {
+		return !reequipO.get();
```
BluSpring: Make sure to check over the original boolean too, this won't stack with other mixins without it.

src/main/java/xyz/bluspring/kilt/forgeinjects/client/renderer/ItemInHandRendererInject.java
```diff
@@ -62,28 +64,24 @@ public abstract class ItemInHandRendererInject {
...
-    // TODO: implement when ternaries are fixed in MixinExtras
-    /*@Definition(id = "mainHandItem", field = "Lnet/minecraft/client/renderer/ItemInHandRenderer;mainHandItem:Lnet/minecraft/world/item/ItemStack;")
-    @Definition(id = "itemStack", local = @Local)
+    @Definition(id = "mainHandItem", field = "Lnet/minecraft/client/renderer/ItemInHandRenderer;mainHandItem:Lnet/minecraft/world/item/ItemStack;")
+    @Definition(id = "itemStack", local = @Local(type = ItemStack.class, ordinal = 0))
     @Expression("this.mainHandItem == itemStack")
     @ModifyExpressionValue(method = "tick", at = @At("MIXINEXTRAS:EXPRESSION"))
-    private boolean kilt$useReequipCheckForMainHand(boolean original) {
+    private boolean kilt$useReequipCheckForMainHand(boolean original, @Share("reequipM") LocalBooleanRef reequipM) {
+		return !reequipM.get();
```
BluSpring: Make sure to check over the original boolean too, this won't stack with other mixins without it.

src/main/java/xyz/bluspring/kilt/forgeinjects/client/renderer/ItemInHandRendererInject.java
```diff
@@ -62,28 +64,24 @@ public abstract class ItemInHandRendererInject {
             this.offHandItem = offhandStack;
     }
 
-    // TODO: implement when ternaries are fixed in MixinExtras
-    /*@Definition(id = "mainHandItem", field = "Lnet/minecraft/client/renderer/ItemInHandRenderer;mainHandItem:Lnet/minecraft/world/item/ItemStack;")
-    @Definition(id = "itemStack", local = @Local)
```
BluSpring, line 67: okay this isn't a review on your part but wtf I tried doing this previously? and it didn't work??? wait what??? past me you good??????

BluSpring (changes requested):

src/main/java/xyz/bluspring/kilt/forgeinjects/client/renderer/ItemInHandRendererInject.java
Acuadragon100: About that, is there a way to detect if an item comes from a forge mod or a fabric mod? Because then we could prefer original when it's a fabric item and prefer the forge reequip check when it's a forge item, or is that a bad idea?

BluSpring: Technically there are two ways, however this is being run in rendering code which is a significantly more hot path than ticking code, so by doing so we would likely end up introducing a performance penalty so I don't think it's entirely worth it.

However, if you do want to try it out and see if it does introduce a significant performance penalty, what I would usually end up doing is I would use `KiltHelper.INSTANCE.hasMethodOverride`, I would check if there is an override for `shouldCauseReequipAnimation` on at least one of the two items, and if there is, then it would enforce the Forge version while still stacking with the original boolean.

Acuadragon100: Ah, thanks! Yeah, that function goes through a lock to access the cache, so maybe not the best idea here. I tried adding another cache in ItemInHandRendererInject using an Object2BooleanMap but that still seemed several times slower on my machine, on top of the extra memory requirement that brings.

So unless we take the performance hit, I think we're going to have to choose to give priority to either all forge or all fabric mods. We can either allow both platforms to skip the animation or to trigger it, the other behaviour will only work on the platform we give priority to. I think it might make more sense to give both platforms the ability to skip the animation. As for which platform should get priority, I'm not so sure. On the one hand it makes sense to choose fabric, but I don't know how many fabric mods actually use this exact mixin location. Some of them might put the mixin in a different location to achieve the same effect. Forge mods on the other hand are probably more likely to rely on IForgeItem::shouldCauseReequipAnimation. But what do you think? Which platform do you think should get preference? Do you have another suggestion?

BluSpring: Personally I would rather Fabric have priority over Forge, Fabric mods are usually a lot more of a wildcard.

BluSpring (approved):
I like this a lot better, awesome, thank you

BluSpring:
Okay, for Forge mod tests, I recommend looking into Mekanism (energized items seem to use it) and whatever else is available on 1.20.1 with [this](https://github.com/search?q=shouldCauseReequipAnimation&type=code)

Acuadragon100:
I tried a few and the equip animation did not seem to be an issue with any of them.

PR #529, opened 2025-10-24
Register TransferInterop to Forge event bus correctly

KorewaLidesu:
The event register before was not fired due to incorrect registration.
Should have register via object init instead of method.

### DIFF
--- a/compat/transfer-api-compat/src/main/kotlin/xyz/bluspring/kilt/compat/transfer/TransferInterop.kt
+++ b/compat/transfer-api-compat/src/main/kotlin/xyz/bluspring/kilt/compat/transfer/TransferInterop.kt
@@ -26,7 +26,7 @@ import xyz.bluspring.kilt.compat.transfer.item.ForgeSlottedStorage
 
 class TransferInterop : ModInitializer {
     override fun onInitialize() {
-        MinecraftForge.EVENT_BUS.register(TransferInterop())
+        MinecraftForge.EVENT_BUS.register(this)
 
         ItemStorage.SIDED.registerFallback { world, pos, state, blockEntity, direction ->
             if (blockEntity == null)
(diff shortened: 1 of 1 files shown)


compat/transfer-api-compat/src/main/kotlin/xyz/bluspring/kilt/compat/transfer/TransferInterop.kt
```diff
@@ -26,8 +26,7 @@ import xyz.bluspring.kilt.compat.transfer.item.ForgeSlottedStorage
 
 class TransferInterop : ModInitializer {
     override fun onInitialize() {
-        MinecraftForge.EVENT_BUS.register(::onAttachBlockEntityCapabilities)
-        MinecraftForge.EVENT_BUS.register(::onAttachItemStackCapabilities)
+        MinecraftForge.EVENT_BUS.register(TransferInterop())
```
BluSpring: Use `this` instead of creating a new instance of `TransferInterop`.

Also, does this manage to definitely solve the problem? I swear this worked before but I may be wrong...

BluSpring (changes requested):

compat/transfer-api-compat/src/main/kotlin/xyz/bluspring/kilt/compat/transfer/TransferInterop.kt
KorewaLidesu: Yes, Forge event can be abit deceive when come to properly register event.
`register` is for class-type register, for method should be `addListener` or `addGenericListener`

Edit: You can find it [here](https://docs.minecraftforge.net/en/1.20.x/concepts/events/#instance-annotated-event-handlers)
`To register this event handler, use MinecraftForge.EVENT_BUS.register(...) and pass it an instance of the class the event handler is within. If you want to register this handler to the mod specific event bus, you should use FMLJavaModLoadingContext.get().getModEventBus().register(...) instead.`

BluSpring: Gotcha, alright.
I would definitely still prefer using `this` over creating a new instance entirely though.

KorewaLidesu:
Can you please squash this PR? I merged origin branch by accident.

BluSpring:
Don't worry, I always squash and merge

PR #578, opened 2025-12-09
Fix Tropicraft not working with Registrate-Refabricated.

### LINKED ISSUE #349: Crash on startup when loading Tropicraft with Create Fabric
### Version information

20.1.2+build.c0f636-nightly MC 1.20.1 Fabric 0.6.14

### Reproduction Steps

The issue where Tropicraft doesn't work with Kilt was fixed, but I ran into a new issue. Loading Tropicraft with Kilt and Create Fabric causes a crash. Game runs properly without Create Fabric.

### Java Provider + Version

Arch Linux OpenJDK 24.0.1

### Crash Report file

[crash-2025-07-09_17.54.55-client.txt](https://github.com/user-attachments/files/21153548/crash-2025-07-09_17.54.55-client.txt)

### Additional information

From the logs it looks like a Feature incompatibility issue or a BlockState related issue. Perhaps with the Create patches in Kilt or a breaking change Create Fabric makes? If the latter I'd be willing to submit this issue to Create Fabric as well.

`Caused by: java.lang.ClassCastException: class net.minecraft.class_2189 cannot be cast to class net.tropicraft.core.common.block.CoffeeBushBlock (net.minecraft.class_2189 and net.tropicraft.core.common.block.CoffeeBushBlock are in unnamed module of loader 'knot' @1794d431)
		at knot//net.tropicraft.core.common.dimension.feature.CoffeePlantFeature.<clinit>(CoffeePlantFeature.java:18)`

Acuadragon100:
This pull request fixes various crashes when Tropicraft is used with Registrate-Refabricated.

I decided to add an access widener to the create compatibility module so I could extend EntityType$Builder with a wrapper that calls FabricEntityTypeBuilder instead. Not 100% sure I'm using the mixin-injected function names you prefer to set the Forge-specific properties in the EntityBuilder since Kotlin didn't like dollar signs in the function names, but it seems to run just fine.

Closes #349


compat/create-compat/src/main/kotlin/xyz/bluspring/kilt/compat/create/extensions/EntityBuilderExtension.kt
```diff
@@ -0,0 +1,18 @@
...
+import net.minecraft.world.entity.Entity
+import net.minecraft.world.entity.EntityType
+import net.minecraft.world.level.Level
+import net.minecraftforge.network.PlayMessages
+import java.util.function.BiFunction
+import java.util.function.Predicate
+import java.util.function.ToIntFunction
+
+interface EntityBuilderExtension<T> {
+
```
BluSpring: Usually I just use Java for these types of extensions, either that or at least try to make it look nicer in Kotlin like this:
```kt
@JvmName("kilt\$setCustomClientFactory")
fun setCustomClientFactory(customClientFactory: /* [...] */)
```

Do remove the spacings for the types though, so it looks like L13 instead.

compat/create-compat/src/main/resources/kilt-create-compat.accesswidener
```diff
@@ -0,0 +1,3 @@
+accessWidener v2 named
+
+extendable method net/minecraft/world/entity/EntityType$Builder <init> (Lnet/minecraft/world/entity/EntityType$EntityFactory;Lnet/minecraft/world/entity/MobCategory;)V
```
BluSpring: This should not be extendable, just use acccessible. Unless for whatever reason AW makes the constructor final somehow.

compat/create-compat/src/main/kotlin/xyz/bluspring/kilt/compat/create/EntityBuilderAdapter.kt
```diff
@@ -0,0 +1,85 @@
...
+import net.minecraft.world.level.Level
+import net.minecraft.world.level.block.Block
+import net.minecraftforge.network.PlayMessages.SpawnEntity
+import xyz.bluspring.kilt.compat.create.extensions.EntityBuilderExtension
+import java.util.function.BiFunction
+
+
+class EntityBuilderAdapter<T : Entity>(
+    val builder : FabricEntityTypeBuilder<T>,
+    val registrateBuilder : EntityBuilderExtension<T>
```
BluSpring: Avoid spacing here (`val builder: FabricEntityTypeBuilder<T>`).

compat/create-compat/src/main/java/xyz/bluspring/kilt/compat/create/mixin/registrate_fabric/EntityBuilderMixin.java
```diff
@@ -0,0 +1,101 @@
...
+            at = @At(
+                    value = "INVOKE",
+                    target = "Lcom/tterrag/registrate/util/nullness/NonNullConsumer;andThen(Lcom/tterrag/registrate/util/nullness/NonNullConsumer;)Lcom/tterrag/registrate/util/nullness/NonNullConsumer;"
+            ),
+            remap = false
+    )
+    public NonNullConsumer<Object> properties(NonNullConsumer<Object> after) {
+        if (Kilt.Companion.getLoader().hasMod(getOwner().getModid())) {
+            return fabricBuilder -> after.accept(
+                    new EntityBuilderAdapter<>((FabricEntityTypeBuilder<T>) fabricBuilder, this)
```
BluSpring: I'm not familiar enough with Registrate, is this cast always going to be valid? Why is it the Fabric entity type builder exactly?
Also, be careful, make sure to store a variable reference to `this` outside of the lambda (`EntityBuilder<T, P> self = (EntityBuilder<T, P) (Object) this;`). I've encountered weird issues in the past with mixin where the internal lambda descriptor contains a reference to the mixin class directly rather than renaming to the target class, causing "mixin class loaded directly" exceptions.

BluSpring (changes requested):

compat/create-compat/src/main/java/xyz/bluspring/kilt/compat/create/mixin/registrate_fabric/EntityBuilderMixin.java
Acuadragon100: Basically, Registrate Refabricated uses the FabricEntityTypeBuilder to construct the entity while Forge Registrate uses EntityType$Builder. So all Forge mods calling the `properties` function will implicitly typecast it to EntityType$Builder, which will crash when using Registrate Refabricated.

As you can see, [forge Registrate always uses EntityType$Builder](https://github.com/tterrag1098/Registrate/blob/1.20/src/main/java/com/tterrag/registrate/builders/EntityBuilder.java#L103) while [Registrate Refabricated always uses FabricEntityTypeBuilder](https://github.com/Fabricators-of-Create/Registrate-Refabricated/blob/fabric/1.20/src/main/java/com/tterrag/registrate/builders/EntityBuilder.java#L103). If this is ever not a FabricEntityTypeBuilder, then we're either using Forge Registrate for some reason despite Registrate Fabric being present, or some other mod is doing something at least as hacky as this, and I can't think of any reason why they would do that.

Still, I guess it doesn't hurt to do an instanceof check and just forwarding it as is if it looks invalid, hoping whatever mod is messing with this knows what it's doing.

BluSpring: I wouldn't be surprised if Forge Registrate gets loaded somehow - it doesn't even have a mods.toml, so Kilt can't properly figure out whether Registrate Forge is being loaded or not.

compat/create-compat/src/main/kotlin/xyz/bluspring/kilt/compat/create/extensions/EntityBuilderExtension.kt
Acuadragon100: It didn't seem to like putting the JvmName annotation on interface functions, so I think I'll use a Java interface, but it might just have been skill issue on my part.

BluSpring (approved):
Looks fine to me, will test later when I'm available.

Cartrigger:
it works

BluSpring:
This experience from the Discord needs to be hung up on a wall
image
image
image

PR #589, opened 2025-12-15
fix for GTCEu's LevelMixin colliding with Porting Lib

Brickmaster1:


Brickmaster1:
Still needs some work :sweat_smile: 
https://mclo.gs/e05aQz5

PR #615, opened 2026-01-10
Fix Striders dying in lava when Valkyrien Skies is installed.

Acuadragon100:
I've discovered a major problem. Whenever Valkyrien Skies is installed, the game makes lava behave like water in various ways. One major side effect of this is that Striders die in lava.

This pull request resolves that. First off, [Valkyrien Skies uses cancellable instead of ModifyReturnValue](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/blob/206b5c9eb1ebf25bdd23e998f66dc2957aa6dfc3/fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/water_in_ships_entity/MixinEntity.java#L179) to modify the output of Entity::updateFluidHeightAndDoFluidPushing, causing EntityInject::kilt$ensureIsActuallyInFluidType to miss it. I fixed this by modifying the field Valkyrien Skies returns in the cancellable inject.

However, there is another issue. It invokes Entity::updateFluidHeightAndDoFluidPushing recursively, which means solving the first problem will falsely make fluids in the shipyard return false. I solve this by passing the fluid type and interim calculations found in the shipyard to the calling function via some variables so they can be used instead.

One thing about this patch is that I use reflection to fetch some variables from the mixin manually, which might not be the most elegant way to do it. Let me know if there is a better way. Funnily enough I didn't actually need to include Valkyrien Skies as an explicit dependency in the gradle configuration because of this.

### DIFF
--- a/compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/valkyrienskies/EntityMixin.java
+++ b/compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/valkyrienskies/EntityMixin.java
@@ -3,67 +3,34 @@ package xyz.bluspring.kilt.compat.fabric.mixin.valkyrienskies;
 import com.llamalad7.mixinextras.sugar.Share;
 import com.llamalad7.mixinextras.sugar.ref.LocalRef;
 import com.moulberry.mixinconstraints.annotations.IfModLoaded;
-import io.github.fabricators_of_create.porting_lib.fluids.PortingLibFluids;
 import it.unimi.dsi.fastutil.objects.Object2ObjectMap;
-import net.minecraft.tags.FluidTags;
 import net.minecraft.tags.TagKey;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.level.material.Fluid;
 import net.minecraft.world.phys.AABB;
 import net.minecraft.world.phys.Vec3;
-import net.minecraftforge.common.ForgeMod;
 import net.minecraftforge.fluids.FluidType;
 import org.apache.commons.lang3.tuple.MutableTriple;
+import org.spongepowered.asm.mixin.Dynamic;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.Unique;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
-
-import java.lang.reflect.Field;
+import xyz.bluspring.kilt.Kilt;
 
 @IfModLoaded("valkyrienskies")
-@Mixin(value = Entity.class, priority = 999)
+@Mixin(value = Entity.class, priority = 1101)
 public class EntityMixin {
 
-    @Unique
-    private static final String KILT$ENTITY_INJECT = "xyz.bluspring.kilt.forgeinjects.world.entity.EntityInject";
+    @Dynamic("Valkyrien Skies will return this boolean from updateFluidHeightAndDoFluidPushing and bypass ModifyReturnValue")
+    @Shadow(remap = false)
+    private boolean valkyrienskies$fluidPushRet;
 
-    @Unique
-    private static Field kilt$vsFluidAABBField = null;
-
-    @Unique
-    private static Field kilt$vsFluidPushRetField = null;
-
-    // Field name is dynamically decorated by mixinextras.
-    // This was the best way I could think of to find it reliably.
-    @Unique
-    private static Field kilt$getVSFluidPushRetField() {
-        if (kilt$vsFluidPushRetField == null) {
-            for (var field : Entity.class.getDeclaredFields()) {
-                if (field.getType() == Boolean.TYPE && field.getName().endsWith("valkyrienskies$fluidPushRet")) {
-                    kilt$vsFluidPushRetField = field;
-                    field.setAccessible(true);
-                    break;
-                }
-            }
-        }
-        return kilt$vsFluidPushRetField;
-    }
-
-    @Unique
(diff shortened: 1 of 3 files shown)


compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/valkyrienskies/EntityMixin.java
```diff
@@ -0,0 +1,160 @@
...
+    // VS calls updateFluidHeightAndDoFluidPushing recursively on the shipyard,
+    // we need to pass the fluid type from that one out.
+    @Inject(
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/world/phys/Vec3;length()D", ordinal = 0),
+            method = "updateFluidHeightAndDoFluidPushing"
+    )
+    private void kilt$fixShipyardReturnValue(
+            TagKey<Fluid> fluidTag, double motionScale, CallbackInfoReturnable<Boolean> cir,
+            @Share(value = "fluidType", namespace = KILT$ENTITY_INJECT) LocalRef<FluidType> fluidTypeRef,
+            @Share(value = "interimCalcs", namespace = KILT$ENTITY_INJECT) LocalRef<Object2ObjectMap<FluidType, MutableTriple<Double, Vec3, Integer>>> interimCalcs
```
BluSpring: The namespace should just be "kilt", set that in the main inject. Too verbose and likely prone to breakage.

compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/valkyrienskies/EntityMixin.java
```diff
@@ -0,0 +1,160 @@
...
+            for (var field : Entity.class.getDeclaredFields()) {
+                if (field.getType() == AABB.class && field.getName().endsWith("valkyrienskies$fluidPushAABB")) {
+                    kilt$vsFluidAABBField = field;
+                    field.setAccessible(true);
+                    break;
+                }
+            }
+        }
+        return kilt$vsFluidAABBField;
+    }
```
BluSpring: You might be able to target this with `@Shadow` actually, as long as the priority is higher.

compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/valkyrienskies/EntityMixin.java
```diff
@@ -0,0 +1,160 @@
...
+    // Valkyrien Skies uses cancellable instead of ModifyVariable, so this hack is necessary.
+    // https://github.com/ValkyrienSkies/Valkyrien-Skies-2/blob/206b5c9eb1ebf25bdd23e998f66dc2957aa6dfc3/fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/water_in_ships_entity/MixinEntity.java#L179
+    @Inject(
+            method = "updateFluidHeightAndDoFluidPushing",
+            at = @At(value = "RETURN", ordinal = 1)
+    )
+    private void kilt$fixVSReturnValue(
+            TagKey<Fluid> fluidTag, double motionScale, CallbackInfoReturnable<Boolean> cir,
+            @Share(value = "fluidType", namespace = KILT$ENTITY_INJECT) LocalRef<FluidType> fluidTypeRef,
+            @Share(value = "interimCalcs", namespace = KILT$ENTITY_INJECT) LocalRef<Object2ObjectMap<FluidType, MutableTriple<Double, Vec3, Integer>>> interimCalcs
```
BluSpring: See above.

compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/valkyrienskies/EntityMixin.java
```diff
@@ -0,0 +1,160 @@
...
+
+            kilt$vsFluidType = null;
+            kilt$vsInterimCalcs = null;
+        } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Unique
+    private boolean isCorrectFluidTag(
```
BluSpring: Prefix with `kilt$vs$` or something.

compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/valkyrienskies/EntityMixin.java
```diff
@@ -0,0 +1,160 @@
...
+            else if (interimCalcs != null)
+                return interimCalcs.containsKey(ForgeMod.WATER_TYPE.get());
+        } else if (fluidTag == FluidTags.LAVA) {
+            if (fluidType != null && !fluidType.isAir())
+                return fluidType == ForgeMod.LAVA_TYPE.get() || fluidType == PortingLibFluids.LAVA_TYPE;
+            else if (interimCalcs != null)
+                return interimCalcs.containsKey(ForgeMod.LAVA_TYPE.get());
+        }
+        return defaultReturn;
+    }
```
BluSpring: In theory, wouldn't it be possible for this to be made into a separated call in Kilt's EntityInject instead, and using MixinSquared + `@Shadow` to target it?

BluSpring (changes requested):
I won't lie, the fluid injects in EntityInject is one of my most hated injects I've had to make, because of how terribly incompatible that code is. If you're able to figure out how to make that code more compatible it would be much appreciated, because I really do not like that code.

compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/valkyrienskies/EntityMixin.java
Acuadragon100: Well, it seems to work at least.

BluSpring (approved):

PR #621, opened 2026-01-15
Fixed wither storm rendering issue while using shaders 

Vorty-SH:
**Reproduction steps**

1:Download Kilt
2:Install Iris Shaders and any Shader pack
3:launch minecraft and load any shader
4:Create a world
5:Spawn the wither storm 

using-
kilt build 20.1.13 + Cracker's Wither Storm Mod 1.20.1-4.2.1

Turn on shaders (doesn't matter which one, used photon for my usecase) 

Screenshot 2026-01-15 191046

the wither storm is invisible

**AFTER FIX**

Screenshot 2026-01-15 192116

the witherstorm is visible with all shaders now.
what I did was modify Kilt's code to make witherstorm mod think that shaders are always running via MixIns, thanks to @Digitalgaming1 on github for figuring it out, I just implemented it


Vorty-SH:
Note:
This is my first time contributing on GitHub, so there were a few intermediate commits
where files were temporarily placed incorrectly. These have all been cleaned up.

The final state of the branch contains only the intended changes.
I have built the project locally using ./gradlew remapJar and confirmed that the fix
works as expected.

The PR is ready for review and merging.

BluSpring:
**No.**
Just, no.
This code is wrong. So very wrong.
This mixin is skipping over [an entire shader check](https://github.com/nonamecrackers2/crackerslib/blob/1.20.1/src/main/java/nonamecrackers2/crackerslib/common/compat/CompatHelper.java#L44-L71) to instead short circuit to always return true for shaders being enabled, even if they aren't actually on.

The actual fix is much, much simpler, and doesn't even need mixins. Figure out how Git works properly (not just GitHub), revert your codebase, look into [this code](https://github.com/KiltMC/Kilt/blob/version/1.20.1/src/main/kotlin/net/minecraftforge/fml/ModList.kt), and try again.

Vorty-SH:
sure, I will try my best

Digitalgaming1:
uhh yeah and also AI wrote that code with just a temp fix for one use case which is also why i didn't make a pull request

BluSpring:
Note that I have written a similar system to what is actually wanted [here](https://github.com/KiltMC/Kilt/blob/version/1.20.1/src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoader.kt#L81-L86).

PR #690, opened 2026-03-06
Fix crash with amecs 1.6 (amecsapi 1.8).

Acuadragon100:
The latest version of amecs splits different parts of the codebase into separate modules.
This means that the class paths and class names are different, which means that Kilt's compatibility mixins no longer work.

Luckily, the underlying code is essentially the same, making this an easy fix.
That said, to maintain backwards compatibility I decided to just make a copy of the old code.

Also worth mentioning that KeyBindingManager is now split into AmecsKeyMappingManager and AmecsKeyMappingManagerLayer. KeyBindingManager was client-only before, but now only AmecsKeyMappingManager is client-only. AmecsKeyMappingManagerLayer is common, so I made the mixin common as well. Do you think I should mark the mixin as client-only?

### DIFF
--- a/compat/fabric-compats/build.gradle.kts
+++ b/compat/fabric-compats/build.gradle.kts
@@ -21,6 +21,9 @@ dependencies {
     modCompileOnly("de.siphalor:amecsapi-1.20:${property("amecsapi_version")}") {
         isTransitive = false
     }
+    modCompileOnly("de.siphalor.amecs.amecs-key-modifiers:amecs-key-modifiers-mc1.20.0:${property("amecs_key_modifiers_version")}") {
+        isTransitive = false
+    }
     modCompileOnly("dev.emi:emi-fabric:${property("emi_version")}")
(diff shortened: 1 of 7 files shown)


BluSpring:
I don't think it ultimately really matters, maybe mark the mixin classes as `@Pseudo` first for safety?

Acuadragon100:
Honestly, I don't think adding `@Pseudo` will make any notable difference. In particular, KeyMappingMixin which relies on mixinsquared still seemed to crash without the maxVersion property, even with `@Pseudo`.

That said, even for the cases where `@Pseudo` does avoid a crash, the main benefit would be that it wouldn't crash if there are changes made to Amecs that would require the mixin to be updated. It would mean fewer crashes, but the key modifier mechanic of Amecs would still most likely stop working, and then it might not be as obvious which mod is to blame (or where to start debugging).

Acuadragon100:
Do you want me to add `@Pseudo` anyway?
If so, should I add it to the old amecsapi mixins, the new mixins, or both?

BluSpring:
Both tbh

PR #709, opened 2026-03-25
Allow KiltMixinModifications to use Mojang mappings instead of SRG.

Acuadragon100:
This makes the current modifications specified in KiltMixinModifications work in production without having to translate them to SRG.
It also makes sure annotation replacements are properly remapped (hence why I changed MixinRemapper so the remapAnnotations function is now exposed).

Since there are 2 enhanced remappers I renamed the old one to enhancedSRGRemapper to minimise confusion.
The rest of the changes are mostly putting stuff into functions so we can reuse the same code for SRG and Mojang.

It should be noted that method remapping in KiltMixinModifications still uses SRG primarily and only falls back to Mojang mappings when no SRG name is found, whereas ReplacedAnnotationsModifier and NameRemappingAnnotationModifier only remap to Mojang mappings.

Do you think I should remove the srg mapping from KiltMixinModifications::remapMethod?

### DIFF
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/remap/KiltRemapper.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/remap/KiltRemapper.kt
@@ -96,7 +96,7 @@ object KiltRemapper {
     // srg -> intermediary
     val srgIntermediaryMapping: IMappingFile = mappingFile.getMap("searge", "intermediary")
     // mojang -> intermediary
-    val intermediaryMojMapping: IMappingFile = mappingFile.getMap("intermediary", "mojang").reverse()
+    val mojIntermediaryMapping: IMappingFile = mappingFile.getMap("intermediary", "mojang").reverse()
 
     val fabricMappings: INamedMappingFile = MappingConfiguration::class.java.classLoader.getResourceAsStream("mappings/mappings.tiny")!!.use { INamedMappingFile.load(it) }
 
@@ -107,7 +107,7 @@ object KiltRemapper {
             this
     }
 
-    private val devMojangIntermediaryMapping = intermediaryMojMapping.run {
+    private val devMojangIntermediaryMapping = mojIntermediaryMapping.run {
         if (!forceProductionRemap)
             this.rename(DevMappingRenamer(FabricLoader.getInstance().mappingResolver))
         else
@@ -127,7 +127,7 @@ object KiltRemapper {
         this::class.java.getResourceAsStream("/kilt_workaround_mappings.tiny")!!.bufferedReader()
     )
 
-    lateinit var enhancedSRGRemapper: KiltEnhancedRemapper
+    lateinit var enhancedSrgRemapper: KiltEnhancedRemapper
     lateinit var enhancedMojangRemapper: KiltEnhancedRemapper
 
     private lateinit var remappedModsDir: Path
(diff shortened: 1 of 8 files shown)


BluSpring:
> Do you think I should remove the srg mapping from KiltMixinModifications::remapMethod?

Personally don't think it matters

src/main/kotlin/xyz/bluspring/kilt/loader/remap/KiltRemapper.kt
BluSpring: I'm realizing now this should be renamed to mojIntermediaryMapping

src/main/kotlin/xyz/bluspring/kilt/loader/remap/KiltRemapper.kt
```diff
@@ -119,7 +127,8 @@ object KiltRemapper {
         this::class.java.getResourceAsStream("/kilt_workaround_mappings.tiny")!!.bufferedReader()
     )
 
-    lateinit var enhancedRemapper: KiltEnhancedRemapper
+    lateinit var enhancedSRGRemapper: KiltEnhancedRemapper
```
BluSpring: Personal nitpick - the name should be `enhancedSrgRemapper`

BluSpring (changes requested):
Slightly worried about the memory usage implications but oh well

Acuadragon100:
I get what you mean.
We could always look into remapping at compile time instead, but I'm unsure about the best way to do that.

BluSpring:
Thinking about it, since the remapped values are only ever used in the modifiers and they're remapped ahead of time, the Mojang mappings and remapper could be discarded after the mixin modifiers are fully remapped, and since the Mojang-mapped jar is cached, it doesn't take nearly as much time afterwards.

Acuadragon100:
Ah, we could do that.
That said, NameRemappingAnnotationModifier and ReplacedAnnotationsModifier seem to remap the mixins when they are loaded right now (which might be after startup), so we'd need to move the remapping step to the register function in that case, which would include marking some variables as mutable.

Should I look into that or should I leave it as is?

BluSpring:
Leave it as is, the load order kinda ensures that mixin modifiers are remapped after the remapper is already initialized, and after they're all registered you can probably just mark the variable as null

Acuadragon100:
Hold on... Sorry, I'm not following.
At what point would I set the Mojang remapper to null exactly?
With the way it's currently structured, some mixins aren't remapped until you load (or try to create) a world because that's when the mixin is loaded...

BluSpring:
The mixin modifiers' remapping occurs at the same time as when they are registered. In that case, if you add a static initializer (`init {}` block) after everything in KiltMixinModifications, anything running in that block will run after the modifiers have already been remapped, so you can clear the remappers by that point, in theory at least.

Acuadragon100:
Yeah, but what about this function call?
Wouldn't I need to move that out into the register function to do that?
https://github.com/KiltMC/Kilt/blob/8998acffba421b6b4b5f54ce6a2c830c87425040/src/main/kotlin/xyz/bluspring/kilt/loader/mixin/modifications/modifiers/AnnotationBasedModifier.kt#L69

BluSpring:
If you do something similar to what the `mappedOwner` and `mappedMethods` system are doing by remapping the `remapMethodsTo` field ahead of time, this should be pretty easy to deal with
I don't think I got to doing this cuz I forgot about it

Acuadragon100:
Ah ok, fair enough.
I made an attempt, but I'm not 100% sure about the best way to mark the variables as nullable.

BluSpring:
Looks good enough at least, I can do something about it in the future if I'm not too happy about the nullability

BluSpring (approved):

PR #715, opened 2026-03-31
AutoModpack Support

Acuadragon100:
There are currently 2 issues preventing AutoModpack from working properly with Kilt.
1. Kilt does not load forge mods from AutoModpack's selected modpack directory. This is fixed with an implementation of the Knit service API (depends on https://github.com/KiltMC/KnitLoader/pull/1).
2. Kilt rejects vanilla clients, which prevents AutoModpack from sending modpacks to the client unless Kilt is installed. This is fixed by enabling vanilla clients to connect (depends on https://github.com/KiltMC/NeoForge/pull/10).

### DIFF
--- a/compat/fabric-compats/gradle.properties
+++ b/compat/fabric-compats/gradle.properties
@@ -42,3 +42,6 @@ pehkui_version = 3.8.3+1.14.4-1.21
 
 # https://modrinth.com/mod/architectury-api/versions?g=1.20.1
 architectury_version=9.2.14
+
+# https://modrinth.com/mod/automodpack/versions?g=1.20.1
+automodpack_version=4.0.5-fabric,1.20.1
--- a/compat/fabric-compats/src/main/kotlin/xyz/bluspring/kilt/compat/fabric/automodpack/KiltAutoModpackCompat.kt
+++ b/compat/fabric-compats/src/main/kotlin/xyz/bluspring/kilt/compat/fabric/automodpack/KiltAutoModpackCompat.kt
@@ -1,0 +1,12 @@
+package xyz.bluspring.kilt.compat.fabric.automodpack
+
+import pl.skidam.automodpack_core.GlobalVariables
+import java.nio.file.Path
+
+object KiltAutoModpackCompat {
+
+    fun getModpackDir(): Path? {
+        return GlobalVariables.selectedModpackDir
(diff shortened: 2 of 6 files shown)


BluSpring:
Requesting pull from current main branch of Knit Loader

compat/fabric-compats/gradle.properties
```diff
@@ -41,4 +41,7 @@ invmove_version = 0.9.0+1.20.1-fabric
 pehkui_version = 3.8.3+1.14.4-1.21
 
 # https://modrinth.com/mod/architectury-api/versions?g=1.20.1
-architectury_version=9.2.14
\ No newline at end of file
+architectury_version=9.2.14
+
+# https://modrinth.com/mod/automodpack/versions?g=1.20.1
+automodpack_version=4.0.5-fabric,1.20.1
```
BluSpring, line 47: wait. you can just. do that??

Acuadragon100:
Ah, will do. Was otherwise thinking of pulling both forge and loader at once.

compat/fabric-compats/gradle.properties
Acuadragon100: Yep, it's listed on the Modrinth Maven support page.
https://support.modrinth.com/en/articles/8801191-modrinth-maven#:~:text=core%3A5.0.1%22)%0A%7D-,Maven%20version%20filters,-Many%20projects%20on

compat/fabric-compats/src/main/kotlin/xyz/bluspring/kilt/compat/fabric/automodpack/KiltAutoModpackCompat.kt
```diff
@@ -0,0 +1,12 @@
+package xyz.bluspring.kilt.compat.fabric.automodpack
+
+import pl.skidam.automodpack_core.GlobalVariables
+import java.nio.file.Path
+
+object KiltAutoModpackCompat {
+
+    fun getModpackDir(): Path? {
```
BluSpring: This could be made into a val instead:
```kt
val modpackDir: Path?
  get() = GlobalVariables.selectedModpackDir
```

BluSpring (changes requested):

PR #729, opened 2026-04-14
Search recursively for side-only lambdas.

Acuadragon100:
Currently it looks like Kilt only remaps lambdas found directly inside of `@Environment`-annotated methods, ignoring lambdas inside those lambdas. This is a problem because some mods like [The Aether: Redux](https://github.com/Zepalesque/The-Aether-Redux/blob/1.20.1/src/main/java/net/zepalesque/redux/client/particle/ReduxParticleTypes.java#L146) use nested lambdas  for client-only code. These lambdas don't have the `@Environment`-annotation during the initial search, which means only the first lambda is ever annotated, resulting in a serverside game crash because the Particle class could not be loaded.

This pull request resolves this by searching each found lambda for additional lambdas without the `@Environment`-annotation recursively, which seems to solve the problem.

### DIFF
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/remap/fixers/EnvironmentLambdaFixer.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/remap/fixers/EnvironmentLambdaFixer.kt
@@ -8,7 +8,7 @@ import org.objectweb.asm.tree.AnnotationNode
 import org.objectweb.asm.tree.ClassNode
 import org.objectweb.asm.tree.InvokeDynamicInsnNode
 import org.objectweb.asm.tree.MethodNode
-import xyz.bluspring.kilt.util.KiltHelper
+import xyz.bluspring.kilt.util.KiltHelper.mergedAnnotations
 import java.lang.invoke.LambdaMetafactory
 
 object EnvironmentLambdaFixer {
@@ -16,16 +16,12 @@ object EnvironmentLambdaFixer {
     const val LAMBDA_METHOD_DESCRIPTOR = "(Ljava/lang/invoke/MethodHandles\$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodHandle;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/CallSite;"
     private val ENVIRONMENT_TYPE = Type.getType(Environment::class.java)
 
-    private fun getAnnotations(methodNode: MethodNode): Collection<AnnotationNode> {
-        return KiltHelper.mergeNullableCollections(methodNode.visibleAnnotations, methodNode.invisibleAnnotations)
-    }
-
     private fun lacksEnvAnnotation(annotations: Collection<AnnotationNode>): Boolean {
         return annotations.none { it.desc == ENVIRONMENT_TYPE.descriptor }
     }
 
     private fun lacksEnvAnnotation(methodNode: MethodNode): Boolean {
-        return lacksEnvAnnotation(getAnnotations(methodNode))
+        return lacksEnvAnnotation(methodNode.mergedAnnotations)
     }
 
     private fun markLambdas(
@@ -70,7 +66,7 @@ object EnvironmentLambdaFixer {
         val methodsToMark = mutableMapOf<Pair<String, String>, AnnotationNode>()
 
         for (methodNode in classNode.methods) {
-            val annotations = getAnnotations(methodNode)
+            val annotations = methodNode.mergedAnnotations
             if (lacksEnvAnnotation(annotations))
                 continue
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/xyz/bluspring/kilt/loader/remap/fixers/EnvironmentLambdaFixer.kt
```diff
@@ -15,42 +16,68 @@ object EnvironmentLambdaFixer {
...
-                        if (insnNode.bsmArgs[1] is Handle) {
-                            val lambdaTarget = insnNode.bsmArgs[1] as Handle
-                            if (lambdaTarget.owner == classNode.name) {
-                                methodsToMark[Pair(lambdaTarget.name, lambdaTarget.desc)] = envAnnotation
-                            }
+                if (insnNode.bsmArgs?.size == 3) {
+                    if (insnNode.bsmArgs[1] is Handle) {
+                        val lambdaTarget = insnNode.bsmArgs[1] as Handle
+                        if (lambdaTarget.owner == classNode.name) {
+                            methodsToMark[Pair(lambdaTarget.name, lambdaTarget.desc)] = envAnnotation
```
BluSpring: You can type `lambdaTarget.name to lambdaTarget.desc` instead, it auto-creates a pair. This is something I didn't take advantage of at the time of writing this.

src/main/kotlin/xyz/bluspring/kilt/loader/remap/fixers/EnvironmentLambdaFixer.kt
```diff
@@ -15,42 +16,68 @@ object EnvironmentLambdaFixer {
     const val LAMBDA_METHOD_DESCRIPTOR = "(Ljava/lang/invoke/MethodHandles\$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodHandle;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/CallSite;"
     private val ENVIRONMENT_TYPE = Type.getType(Environment::class.java)
 
-    fun fixClass(classNode: ClassNode) {
-        val methodsToMark = mutableMapOf<Pair<String, String>, AnnotationNode>()
+    private fun getAnnotations(methodNode: MethodNode): Collection<AnnotationNode> {
+        return KiltHelper.mergeNullableCollections(methodNode.visibleAnnotations, methodNode.invisibleAnnotations)
+    }
```
BluSpring: I feel like we do this so often this should be a helper function in one of the Kotlin util classes, should be formatted like this:
```kt
val MethodNode.mergedAnnotations: Collection<AnnotationNode>
  get() = KiltHelper.mergeNullableCollections(this.visibleAnnotations, this.invisibleAnnotations)
```

src/main/kotlin/xyz/bluspring/kilt/loader/remap/fixers/EnvironmentLambdaFixer.kt
```diff
@@ -15,42 +16,68 @@ object EnvironmentLambdaFixer {
...
+
+    fun fixClass(classNode: ClassNode) {
+        val methodsToMark = mutableMapOf<Pair<String, String>, AnnotationNode>()
+
+        for (methodNode in classNode.methods) {
+            val annotations = getAnnotations(methodNode)
+            if (lacksEnvAnnotation(annotations))
+                continue
+
+            val envAnnotation = annotations.first { it.desc == ENVIRONMENT_TYPE.descriptor }
```
BluSpring: To avoid iterating this twice, it might be worth doing this instead:
```kt
val envAnnotation = annotations.firstOrNull { it.desc == ENVIRONMENT_TYPE.descriptor } ?: continue
```

Incredibly minor performance and memory gain, not much at all but it's there.

BluSpring (changes requested):

src/main/kotlin/xyz/bluspring/kilt/loader/remap/fixers/EnvironmentLambdaFixer.kt
Acuadragon100: Right, not really sure where's the best place to put it though.
I'm thinking either KiltHelper or a new RemappingHelper file (which would then be mostly empty). Suggestions?

Acuadragon100: Oh well, I added it to KiltHelper now.

src/main/kotlin/xyz/bluspring/kilt/loader/remap/fixers/EnvironmentLambdaFixer.kt
```diff
@@ -7,50 +7,73 @@ import org.objectweb.asm.Type
...
+        }
+    }
+
+    fun fixClass(classNode: ClassNode) {
+        val methodsToMark = mutableMapOf<Pair<String, String>, AnnotationNode>()
+
+        for (methodNode in classNode.methods) {
+            val annotations = methodNode.mergedAnnotations
+            if (lacksEnvAnnotation(annotations))
+                continue
```
BluSpring: Since the `firstOrNull` is right below, this is now redundant.

BluSpring (changes requested):

PR #732, opened 2026-04-17
Change Srg -> Moj

Cartrigger:
This has been a pet peeve of mine, that especially on startup of Kilt, it would still say SRG on 1.21.1. Unless NeoForge has invented MojSRGMappings, and we use it, I don't think that SRG exists anymore.

This changes all of them to use the "Moj" or "moj" case. 

Small amounts of other tiny changes like changes in punctuation in comments, and commenting out (asking first) for one. On top of that, one change was made for already mapped jars, were it was hard coded to show the name, I change that to $mappingName as in accordance with other references inside that file.


src/main/kotlin/xyz/bluspring/kilt/loader/asm/AccessTransformerLoader.kt
```diff
@@ -305,18 +305,18 @@ object AccessTransformerLoader {
...
 
         // field / method
         if (split.size > 2 && !split[2].startsWith("#")) {
             if (split[2] == "*") { // Handle all of them
-                val classInfo = remapper.getClass(srgClassName).orElse(null)
+                val classInfo = remapper.getClass(mojClassName).orElse(null)
 
                 if (classInfo == null) {
-                    logger.warn("Missing class reference (SRG: $srgClassName, Intermediary: $intermediaryClassName) for access transform, skipping.")
+                    logger.warn("Missing class reference (SRG: $mojClassName, Intermediary: $intermediaryClassName) for access transform, skipping.")
```
BluSpring: missed a spot :]

src/main/kotlin/xyz/bluspring/kilt/loader/asm/AccessTransformerLoader.kt
```diff
@@ -330,10 +330,10 @@ object AccessTransformerLoader {
                 }
             } else if (split[2] == "*()") {
 
-                val classInfo = remapper.getClass(srgClassName).orElse(null)
+                val classInfo = remapper.getClass(mojClassName).orElse(null)
 
                 if (classInfo == null) {
-                    logger.warn("Missing class reference (SRG: $srgClassName, Intermediary: $intermediaryClassName) for access transform, skipping.")
+                    logger.warn("Missing class reference (SRG: $mojClassName, Intermediary: $intermediaryClassName) for access transform, skipping.")
```
BluSpring: :D

BluSpring (changes requested):
bonk

PR #746, opened 2026-04-28
Use forge load order.

Acuadragon100:
Some Forge mods unfortunately rely on the default load order without explicitly declaring their dependencies.

On Forge, this order depends on the order of keys in a hash map, but at the very least it's always the same for a given modpack. On Kilt, the order is effectively random and changes every startup, which means it sometimes works and sometimes doesn't.

This patch changes that to make Kilt consistent with how Forge sorts the mods. Note that this won't mean that mods will be placed in the exact same order as they will on Forge every time. In fact, the main mod combination which motivated this change (Deep Aether + Lost Aether Content) doesn't work when only those mods are loaded in isolation because the hash map is a lot smaller with only those mods compared to how big it is without any other mods on Forge. However, it does work if you add more forge mods.

### DIFF
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoader.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoader.kt
@@ -479,9 +479,26 @@ class KiltLoader : KnitModLoader<ForgeMod>(Kilt.MOD_ID, "Forge") {
         }
     }
 
+    /**
+     * The natural order of Forge mods (if no dependencies are marked) is whichever order they get placed in a hash map.
+     * See: https://github.com/MinecraftForge/MinecraftForge/blob/1.20.1/fmlloader/src/main/java/net/minecraftforge/fml/loading/UniqueModListBuilder.java#L42
+     * We want to use this order since some Forge mods might not declare their dependencies properly and just blindly rely on the default sorting.
+     */
+    fun getForgeNaturalOrder(): Comparator<ForgeMod> {
+        val modOrder = hashMapOf<String, Int>()
+        for (mod in this.mods) {
+            modOrder[mod.modId] = 0
+        }
+        var index = 0
+        for (modEntry in modOrder) {
+            modEntry.setValue(index++)
+        }
+        return Comparator.comparing { modOrder[it.modId] ?: 0 }
+    }
+
     override fun finishModScanning() {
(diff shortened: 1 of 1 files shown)


BluSpring:
In theory, wouldn't sorting by hash code return a similar sort, or am I wrong on this?

Acuadragon100:
I did try that at first, and in the case of Deep Aether + Lost Aether Content it actually worked better.
Thing is, the hash key might be larger than the highest index used by the internal array in the hash map. So it needs to convert it to an index within the array, which means the actual order depends on the size of the map.

So basically, sorting by hash keys directly might work in general, but there will probably be edge cases where it won't work. That's technically true with this implementation as well, but it also gives users the ability to add or remove unrelated mods to get around the issue, which is what they would have to do on Forge if this issue were to appear there.

BluSpring:
Personally slightly unsure with this approach, especially considering potential drawbacks. Should probably find a way to directly replicate Forge's load order without directly copying the way they load mods, especially considering the edge case you mentioned of loading mods in isolation, which some users very well could be doing.

Acuadragon100:
Thing is, I'm not sure how else to do it. Ultimately the main reason the order isn't the same when the mods are loaded in isolation comes from the fact that Kilt has fewer builtin mods. Without this patch the order will be different every time the game starts, which is not ideal for modpacks. Using hash keys to sort the list isn't ideal either because it might get the order wrong in some edge cases and then users won't be able to change it the order. I guess using hash keys to sort it could be fine if we just add some config that users can use to manually override the load order?

Ignore my previous message btw, I misunderstood your message (sorry about that).

BluSpring:
Honestly considering Kilt now actually has a config people can use, it could be viable to have config-based mod load order, but then comes the problem of "how do we trust that users aren't going to screw this up?"

Acuadragon100:
Thinking about it, Kilt could probably benefit from something like [Dependency Overrides](https://wiki.fabricmc.net/tutorial:dependency_overrides). I'm thinking a configurable load order would fit into that quite nicely.

Acuadragon100:
Now that #765 is merged, users can always define a different order if necessary to make a certain mod work, so I switched to using the hash keys for sorting. It should resolve the issue with Deep Aether and Lost Aether Content even with only a few mods without requiring an explicit dependency override. If there are any mod combinations that don't work with this without a dependency override, they probably wouldn't work reliably on Forge anyway.

PR #747, opened 2026-04-28
Delay the initial setScreen call.

Acuadragon100:
Some Forge mods like Aether and Aether addons like to replace the menu screen using the opening screen event.

Only problem is that on Fabric the screen is set before the mods are initialized, which means they never receive the first screen opening event. On Forge, the game is patched to move this code into the end of `method_29338` (the lambda inside the constructor of Minecraft that runs after the loading overlay is done). So I basically just replicated this with mixins that store the setScreen call in a variable until the forge mods have been initialized and runs it then. I specifically chose to put this in a separate mixin with a lower priority so that any Fabric mods wrapping the setScreen function calls with `@WrapOperation` will still run at the original time.

### DIFF
--- a/src/main/java/xyz/bluspring/kilt/forgeinjects/client/MinecraftInject.java
+++ b/src/main/java/xyz/bluspring/kilt/forgeinjects/client/MinecraftInject.java
@@ -12,7 +12,6 @@ import com.llamalad7.mixinextras.sugar.Local;
 import com.llamalad7.mixinextras.sugar.Share;
 import com.llamalad7.mixinextras.sugar.ref.LocalBooleanRef;
 import com.llamalad7.mixinextras.sugar.ref.LocalRef;
-import com.mojang.realmsclient.client.RealmsClient;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.Options;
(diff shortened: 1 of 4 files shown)


src/main/java/xyz/bluspring/kilt/forgeinjects/client/MinecraftInject.java
```diff
@@ -367,4 +369,52 @@ private void rightClickAir(CallbackInfo ci, @Local ItemStack stack, @Local Inter
     public ForgeGui kilt$getForgeGui() {
         return (ForgeGui) this.kilt$forgeGui;
     }
+
+    // Use lower priority for screen delay. This gives priority to Fabric mods,
+    // allowing them to set their menu screen without interference from the Forge event.
+    @Mixin(value = Minecraft.class, priority = 500)
+    public static class Early {
```
BluSpring: I would prefer for this to be brought into Kilt's own mixin classes rather than the injects package. Prefer the name to stay as `MinecraftMixin`, following a convention for the mixin package, maybe under `mixin.workarounds.early_screen_delay`?

BluSpring (changes requested):

PR #765, opened 2026-05-16
Added Kilt Dependency Overrides

Acuadragon100:
This PR adds a new option to the `kilt_overrides.json` config file: `dependency_overrides`.
It is more or less equivalent to [Fabric Dependency Overrides](https://wiki.fabricmc.net/tutorial:dependency_overrides), but also allows the user to change the dependency ordering.

At the moment, it allows the user to override the `version` (which overrides the constraint field, I felt "version" would be more readable to the user), `type` (optional vs required, but supports incompatible and discouraged as well), `side` (client, server or both) and `ordering` (before, after or none). All fields are optional and will only replace the mod's builtin definition if specified. If no builtin dependency definition is found it default to any version, optional type, both sides and none ordering respectively when a field is unspecified.

One thing worth noting is that it does not include a way to remove the dependency definitions from mods, but simply overriding `type` to `optional` should accomplish the same thing (as well as maybe setting `ordering` to `none` if necessary).

Also, the enum codecs throw an exception when parsing an invalid name. This is by design to help modpack makers find spelling mistakes more quickly.

### DIFF
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoader.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoader.kt
@@ -60,7 +60,6 @@ import xyz.bluspring.knit.loader.mod.ModEnvironment
 import xyz.bluspring.knit.loader.mod.VersionConstraint
 import xyz.bluspring.knit.loader.util.*
 import java.nio.file.Path
-import java.nio.file.Paths
 import java.nio.file.StandardOpenOption
 import java.util.concurrent.ConcurrentHashMap
 import java.util.jar.JarFile
@@ -374,27 +373,25 @@ class KiltLoader : KnitModLoader<ForgeMod>(Kilt.MOD_ID, "Forge") {
             val dependencyOverrides = config.dependencyOverrides[modId]
 
             if (dependencyOverrides != null) {
-                val newDependencies = mutableMapOf<String, ModDependency>()
+                val modifiedDependencies = mutableMapOf<String, ModDependency>()
                 for (dep in dependencies) {
-                    newDependencies[dep.id] = dep
+                    modifiedDependencies[dep.id] = dep
                 }
                 for (dep in dependencyOverrides) {
-                    val prevDep = newDependencies[dep.key]
+                    val prevDep = modifiedDependencies[dep.key]
                     val additionalData = prevDep?.additionalData ?: mapOf()
-                    newDependencies[dep.key] = ModDependency(
+                    modifiedDependencies[dep.key] = ModDependency(
                         id = prevDep?.id ?: dep.key,
                         constraint = dep.value.version.map { ForgeVersionConstraint(it) as VersionConstraint }.orElse(prevDep?.constraint ?: ForgeVersionConstraint(VersionRange.createFromVersionSpec("(,1.0],[1.0,)"))),
                         type = dep.value.type.orElse(prevDep?.type ?: ModDependency.Type.OPTIONAL),
                         side = dep.value.side.orElse(prevDep?.side ?: ModEnvironment.BOTH),
-                        additionalData = if (dep.value.ordering.isPresent) {
-                            additionalData.plus(Pair("ordering", dep.value.ordering.get()))
-                        } else {
-                            additionalData
-                        }
+                        additionalData = dep.value.ordering.map {
+                            additionalData + ("ordering" to it)
+                        }.orElse(additionalData)
                     )
                 }
                 dependencies.clear()
-                dependencies.addAll(newDependencies.values)
+                dependencies.addAll(modifiedDependencies.values)
             }
 
             val definition = ModDefinition(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoaderConfig.kt
```diff
@@ -3,24 +3,65 @@ package xyz.bluspring.kilt.loader
...
+                { VersionRange.createFromVersionSpec(it) },
+                {it.toString()}
+            )
+            val CODEC: Codec<ModDependencyOverride> = RecordCodecBuilder.create { instance ->
+                instance.group(
+                    VERSION_CONSTRAINT_CODEC.optionalFieldOf("version").forGetter { it.version },
+                    EnumUtils.createThrowingFallbackCodec<Type>().optionalFieldOf("type").forGetter { it.type },
+                    EnumUtils.createThrowingFallbackCodec<ModEnvironment>().optionalFieldOf("side").forGetter { it.side },
+                    EnumUtils.createThrowingFallbackCodec<IModInfo.Ordering>().optionalFieldOf("ordering").forGetter { it.ordering }
+                ).apply(instance, ::ModDependencyOverride)
```
BluSpring: Use `.forGetter(ModDependencyOverride::version)` and equivalent for all. Also newline the `forGetter`s and `apply`.

src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoaderConfig.kt
```diff
@@ -3,24 +3,65 @@ package xyz.bluspring.kilt.loader
...
+        val version: Optional<VersionRange> = Optional.empty(),
+        val type: Optional<Type> = Optional.empty(),
+        val side: Optional<ModEnvironment> = Optional.empty(),
+        val ordering: Optional<IModInfo.Ordering> = Optional.empty()
+    ) {
+        companion object {
+            val VERSION_CONSTRAINT_CODEC: Codec<VersionRange> = Codec.STRING.xmap(
+                { VersionRange.createFromVersionSpec(it) },
+                {it.toString()}
+            )
```
BluSpring, line 57: `VersionRange::createFromVersionSpec` and `VersionRange::toString`

src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoader.kt
```diff
@@ -369,6 +371,32 @@ class KiltLoader : KnitModLoader<ForgeMod>(Kilt.MOD_ID, "Forge") {
...
+                    newDependencies[dep.key] = ModDependency(
+                        id = prevDep?.id ?: dep.key,
+                        constraint = dep.value.version.map { ForgeVersionConstraint(it) as VersionConstraint }.orElse(prevDep?.constraint ?: ForgeVersionConstraint(VersionRange.createFromVersionSpec("(,1.0],[1.0,)"))),
+                        type = dep.value.type.orElse(prevDep?.type ?: ModDependency.Type.OPTIONAL),
+                        side = dep.value.side.orElse(prevDep?.side ?: ModEnvironment.BOTH),
+                        additionalData = if (dep.value.ordering.isPresent) {
+                            additionalData.plus(Pair("ordering", dep.value.ordering.get()))
+                        } else {
+                            additionalData
+                        }
```
BluSpring:
```kt
dep.value.ordering.map {
  additionalData + ("ordering" to it)
}.orElse(additionalData)
```

src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoader.kt
```diff
@@ -369,6 +371,32 @@ class KiltLoader : KnitModLoader<ForgeMod>(Kilt.MOD_ID, "Forge") {
...
 
+            val dependencyOverrides = config.dependencyOverrides[modId]
+
+            if (dependencyOverrides != null) {
+                val newDependencies = mutableMapOf<String, ModDependency>()
+                for (dep in dependencies) {
+                    newDependencies[dep.id] = dep
+                }
+                for (dep in dependencyOverrides) {
+                    val prevDep = newDependencies[dep.key]
```
BluSpring: The naming convention feels weird... "previous dependency" coming from `newDependencies` feels wonk
Maybe `newDependencies` should be called `modifiedDependencies`.

BluSpring: The naming convention feels weird... "previous dependency" coming from `newDependencies` feels wonk
Maybe `newDependencies` should be called `modifiedDependencies`.

src/main/kotlin/xyz/bluspring/kilt/loader/KiltLoaderConfig.kt
```diff
@@ -3,24 +3,65 @@ package xyz.bluspring.kilt.loader
...
                     .optionalFieldOf("force_disabled_mods", emptyList())
-                    .forGetter(KiltLoaderConfig::forceDisabledModIds)
+                    .forGetter(KiltLoaderConfig::forceDisabledModIds),
+                    Codec.unboundedMap(
+                        Codec.STRING,
+                        Codec.unboundedMap(
+                            Codec.STRING,
+                            ModDependencyOverride.CODEC
+                        )
+                    ).optionalFieldOf("dependency_overrides", emptyMap())
```
BluSpring: Move these lines back by an indent. I'll have to add in a .editorconfig to Kilt's repo at some point.
Also newline the `optionalFieldOf`.

BluSpring (changes requested):
I'm assuming the format is along the lines of this:
```json
{
  "dependency_overrides": {
    "forge_mod_id": {
      "dependency_id": {
        "type": "optional"
      }
    }
  }
}
```

Acuadragon100:
Yep, that's more or less exactly how it looks like.

BluSpring (approved):

PR #774, opened 2026-05-18
Fix seeds not placeable on rich soil in Farmer's Delight Refabricated.

Acuadragon100:
Rewrites BushBlockInject to allow Fabric mods (such as Farmer's Delight Refabricated) to change the return value. I set the priority to 950 as well to make sure Fabric mods that use `@WrapOperation` around `mayPlaceOn` should wrap around this mixin.

This should still work with the Forge version of Farmer's Delight.

### DIFF
--- a/src/main/java/xyz/bluspring/kilt/forgeinjects/world/level/block/BushBlockInject.java
+++ b/src/main/java/xyz/bluspring/kilt/forgeinjects/world/level/block/BushBlockInject.java
@@ -15,7 +15,7 @@ import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import xyz.bluspring.kilt.util.KiltHelper;
 
-@Mixin(value = BushBlock.class, priority = 950)
+@Mixin(BushBlock.class)
 public abstract class BushBlockInject extends Block implements IPlantable {
     public BushBlockInject(Properties properties) {
         super(properties);
--- a/src/main/java/xyz/bluspring/kilt/forgeinjects/world/level/block/BushBlockInject.java
+++ b/src/main/java/xyz/bluspring/kilt/forgeinjects/world/level/block/BushBlockInject.java
@@ -1,30 +1,37 @@
 // TRACKED HASH: d335200a23b4a1e738aa9a7ea4e390f5f72a6c5d
 package xyz.bluspring.kilt.forgeinjects.world.level.block;
 
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import com.llamalad7.mixinextras.sugar.Local;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
-import net.minecraft.world.level.LevelReader;
+import net.minecraft.world.level.BlockGetter;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.BushBlock;
 import net.minecraft.world.level.block.state.BlockState;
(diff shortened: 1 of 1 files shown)


src/main/java/xyz/bluspring/kilt/forgeinjects/world/level/block/BushBlockInject.java
```diff
@@ -1,30 +1,37 @@
...
-    private void kilt$checkCanSustain(BlockState state, LevelReader level, BlockPos pos, CallbackInfoReturnable<Boolean> cir, @Local(ordinal = 1) BlockPos pos2) {
+    @WrapOperation(
+            method = "canSurvive",
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/world/level/block/BushBlock;mayPlaceOn(Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/BlockGetter;Lnet/minecraft/core/BlockPos;)Z")
+    )
+    private boolean kilt$checkCanSustain(
+            BushBlock instance, BlockState stateAtPos, BlockGetter level, BlockPos posBelow, Operation<Boolean> original,
+            @Local(argsOnly = true) BlockState state
+    ) {
         if (state.getBlock() == this) {
```
BluSpring: This inject predates `KiltHelper.INSTANCE.hasMethodOverride`, this inject should really use that instead.

BluSpring (changes requested):

src/main/java/xyz/bluspring/kilt/forgeinjects/world/level/block/BushBlockInject.java
Acuadragon100: Ah, you mean I should check if there is a method override for `mayPlaceOn`?
I could do that, but I'm pretty sure it's not needed because [BlockInject will end up running that either way](https://github.com/KiltMC/Kilt/blob/version/1.20.1/src/main/java/xyz/bluspring/kilt/forgeinjects/world/level/block/BlockInject.java#L108).

Acuadragon100: Actually, a method override check for `canSustainPlant` would probably make sense.

src/main/java/xyz/bluspring/kilt/forgeinjects/world/level/block/BushBlockInject.java
```diff
@@ -1,30 +1,44 @@
...
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraftforge.common.IPlantable;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
-import org.spongepowered.asm.mixin.injection.Inject;
-import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import xyz.bluspring.kilt.util.KiltHelper;
 
-@Mixin(BushBlock.class)
+@Mixin(value = BushBlock.class, priority = 950)
```
BluSpring: This is already the default priority for all Kilt injects.

BluSpring (changes requested):
There's another change I need to look at later when I'm more awake, namely with the `Block.class` - normally, the hasMethodOverride 2nd param is for the "latest class or interface that implements this method", so if the method owner reference is owned by that class it won't pass. For example, if the latest method variant of param 1's class is implemented by BushBlock overriding IForgeBlock, param 2 should be set to BushBlock.class, otherwise iForgeBlock.class. A lot of hasMethodOverride calls I have are not designed like this in 1.20 since it was still new, but should mostly follow this pattern in 1
21.1.

PR #776, opened 2026-05-22
Rework injects in preparation for NeoForge enum extensions.

Acuadragon100:
I went through all the injects to make sure they would be ready for enum extensions.
All of them appear to work with my enum extension draft (they don't crash at least).

This also removes the older static create methods because they don't seem to be present on NeoForge anyway.

Also, how would you feel about removing RecipeBookCategoriesInjection and RecipeBookTypeInjection?
RecipeBookCategoriesInjection doesn't seem to be used for anything anywhere, but RecipeBookTypeInjection is used to invoke getExtensionInfo from the forge module, although I'm unsure if that will return the correct value.

Edit: Now depends on https://github.com/KiltMC/NeoForge/pull/12

PR #777, opened 2026-05-22
Enum Extension Support

Acuadragon100:
This is the main logic for handling enum extensions. It depends on #776 to avoid crashing when mods try to use it.
It leaves the heavy lifting to Fabric ASM, and makes some smaller manual changes like rewriting `getExtensionInfo` and fixing the index of enums annotated as `@IndexedEnum`. For dealing with `@IndexedEnum`, I am currently just ignoring whatever parameter is provided and instead just forcefully set it to the same value as the ordinal parameter. This should hopefully help avoid issues if Fabric mods just insert a random value and hope for the best since we don't really know what order the enums will end up in `<clinit>`. It also injects initializers into <clinit> of the enum in question that initialize any EnumProxy instances that were referenced in the enum extensions.

Note that due to some enums relying on the integer id for networking, we might want to merge #773 alongside this to make sure the client and the server get the same load order.

The main drawback at the moment is that the Fabric ASM version we're using might run into re-entrance errors in some cases. So I'm marking it as a draft just in there are any mods that seem to work fine without enum extensions but would crash due to re-entrance because of this.

### DIFF
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
@@ -18,7 +18,7 @@ import java.util.function.Consumer
 
 object EnumExtensionLoader {
 
-    val enumExtensions = mutableMapOf<String, MutableSet<EnumPrototype>>()
+    val enumExtensions = mutableSetOf<EnumPrototype>()
     val proxies = mutableMapOf<String, MutableMap<String, EnumParameters.FieldReference>>()
     val indexed = mutableMapOf<String, Int>()
 
@@ -40,10 +40,7 @@ object EnumExtensionLoader {
                 if (Files.isRegularFile(path)) {
                     EnumPrototype.load(mod, path).forEach { prototype ->
                         val usablePrototype = remapPrototype(prototype)
-                        enumExtensions.computeIfAbsent(
-                            usablePrototype.enumName,
-                            { TreeSet() }
-                        ).add(usablePrototype)
+                        enumExtensions.add(usablePrototype)
                     }
                 }
             }
@@ -95,45 +92,43 @@ object EnumExtensionLoader {
     }
 
     fun applyEnumExtensions() {
-        enumExtensions.forEach { (name, prototypes) ->
-            for (prototype in prototypes) {
-                val descriptor = Type.getType(prototype.ctorDesc)
-                val parameters = prototype.ctorParams
-                if (parameters is EnumParameters.FieldReference) {
-                    proxies.computeIfAbsent(prototype.enumName, {mutableMapOf()})[prototype.fieldName] = parameters
-                }
-                ClassTinkerers.enumBuilder(
-                    prototype.enumName, *descriptor.argumentTypes
-                ).addEnum(
-                    prototype.fieldName, {
-                        when (parameters) {
-                            is EnumParameters.Constant -> parameters.params
-                            is EnumParameters.FieldReference -> {
-                                val proxy = getEnumProxy(parameters)
-                                val args = mutableListOf<Any?>()
-                                for (i in 0 until descriptor.argumentCount) {
-                                    args.add(proxy.getParameter(i))
-                                }
-                                args
-                            }
-                            is EnumParameters.MethodReference -> {
-                                // Luckily for us, NeoForge crashes unless these are public.
-                                // This is good because getDeclaredMethod might crash the game on servers.
-                                val method = Class.forName(parameters.owner.className).getMethod(parameters.methodName, Integer.TYPE, Class::class.java)
-                                val args = mutableListOf<Any?>()
-                                for (i in 0 until descriptor.argumentCount) {
-                                    if (indexed[name] == i) {
-                                        args.add(-1)
-                                    } else {
-                                        args.add(method.invoke(null, i, getCastClass(descriptor.argumentTypes[i])))
-                                    }
-                                }
-                                args
-                            }
-                        }.toTypedArray()
-                    }
-                ).build()
+        for (prototype in enumExtensions) {
+            val descriptor = Type.getType(prototype.ctorDesc)
+            val parameters = prototype.ctorParams
+            if (parameters is EnumParameters.FieldReference) {
+                proxies.computeIfAbsent(prototype.enumName, {mutableMapOf()})[prototype.fieldName] = parameters
             }
+            ClassTinkerers.enumBuilder(
+                prototype.enumName, *descriptor.argumentTypes
+            ).addEnum(
+                prototype.fieldName, {
+                    when (parameters) {
+                        is EnumParameters.Constant -> parameters.params
+                        is EnumParameters.FieldReference -> {
+                            val proxy = getEnumProxy(parameters)
+                            val args = mutableListOf<Any?>()
+                            for (i in 0 until descriptor.argumentCount) {
+                                args.add(proxy.getParameter(i))
+                            }
+                            args
+                        }
+                        is EnumParameters.MethodReference -> {
+                            // Luckily for us, NeoForge crashes unless these are public.
+                            // This is good because getDeclaredMethod might crash the game on servers.
+                            val method = Class.forName(parameters.owner.className).getMethod(parameters.methodName, Integer.TYPE, Class::class.java)
+                            val args = mutableListOf<Any?>()
+                            for (i in 0 until descriptor.argumentCount) {
+                                if (indexed[prototype.enumName] == i) {
+                                    args.add(-1)
+                                } else {
+                                    args.add(method.invoke(null, i, getCastClass(descriptor.argumentTypes[i])))
+                                }
+                            }
+                            args
+                        }
+                    }.toTypedArray()
+                }
+            ).build()
         }
     }
 
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
@@ -107,7 +107,7 @@ object EnumExtensionLoader {
                             is EnumParameters.Constant -> parameters.params
                             is EnumParameters.FieldReference -> {
                                 val proxy = getEnumProxy(parameters)
-                                val args = mutableListOf<Any>()
+                                val args = mutableListOf<Any?>()
                                 for (i in 0..< descriptor.argumentCount) {
                                     args.add(proxy.getParameter(i))
                                 }
@@ -117,7 +117,7 @@ object EnumExtensionLoader {
                                 // Luckily for us, NeoForge crashes unless these are public.
                                 // This is good because getDeclaredMethod might crash the game on servers.
                                 val method = Class.forName(parameters.owner.className).getMethod(parameters.methodName, Integer.TYPE, Class::class.java)
-                                val args = mutableListOf<Any>()
+                                val args = mutableListOf<Any?>()
                                 for (i in 0..< descriptor.argumentCount) {
                                     if (indexed[name] == i) {
                                         args.add(-1)
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
@@ -222,10 +222,20 @@ object EnumExtensionLoader {
                     newIndex++
                 }
 
+                val valuesName = KiltRemapper.enhancedRemapper.mapFieldName(
+                    KiltRemapper.unmapClass(targetClass.name),
+                    $$"$VALUES",
+                    "[L${KiltRemapper.unmapClass(targetClass.name)};"
+                )
+
                 // Sort the $VALUES array based on the ordinal after it has been set for the last time.
                 val fixValuesOrder = InsnList()
                 fixValuesOrder.add(LabelNode())
-                fixValuesOrder.add(FieldInsnNode(Opcodes.GETSTATIC, targetClass.name, $$"$VALUES", "[L${targetClass.name};"))
+                fixValuesOrder.add(FieldInsnNode(
+                    Opcodes.GETSTATIC, targetClass.name,
+                    valuesName,
+                    "[L${targetClass.name};"
+                ))
 
                 fixValuesOrder.add(InvokeDynamicInsnNode(
                     "apply", "()Ljava/util/function/Function;",
@@ -244,7 +254,7 @@ object EnumExtensionLoader {
 
                 var setValuesIndex = 0
                 for (ins in clinit.instructions) {
-                    if (ins is FieldInsnNode && ins.name == $$"$VALUES" && ins.owner == targetClass.name && ins.desc == "[L${targetClass.name};" && ins.opcode == Opcodes.PUTSTATIC) {
+                    if (ins is FieldInsnNode && ins.name == valuesName && ins.owner == targetClass.name && ins.desc == "[L${targetClass.name};" && ins.opcode == Opcodes.PUTSTATIC) {
                         setValuesIndex = clinit.instructions.indexOf(ins)
                     }
                 }
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
@@ -23,13 +23,12 @@ object EnumExtensionLoader {
     val indexed = mutableMapOf<String, Int>()
 
     private fun remapPrototype(prototype: EnumPrototype): EnumPrototype {
-        if (FabricLoader.getInstance().isDevelopmentEnvironment) return prototype
         return EnumPrototype(
             prototype.owningMod,
-            KiltRemapper.remapClass(prototype.enumName, toIntermediary = true),
+            KiltRemapper.remapClass(prototype.enumName, toIntermediary = !FabricLoader.getInstance().isDevelopmentEnvironment),
             prototype.fieldName,
-            KiltRemapper.remapDescriptor(prototype.ctorDesc, toIntermediary = true),
-            KiltRemapper.remapDescriptor(prototype.fullCtorDesc, toIntermediary = true),
+            KiltRemapper.remapDescriptor(prototype.ctorDesc, toIntermediary = !FabricLoader.getInstance().isDevelopmentEnvironment),
+            KiltRemapper.remapDescriptor(prototype.fullCtorDesc, toIntermediary = !FabricLoader.getInstance().isDevelopmentEnvironment),
             prototype.ctorParams
         )
     }
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
@@ -2,7 +2,6 @@ package xyz.bluspring.kilt.loader.asm
 
 import it.unimi.dsi.fastutil.objects.Object2IntMap
 import it.unimi.dsi.fastutil.objects.Object2IntOpenHashMap
-import net.fabricmc.loader.api.FabricLoader
 import net.neoforged.fml.common.asm.enumextension.*
 import net.neoforged.fml.common.asm.enumextension.NetworkedEnum.NetworkCheck
 import org.objectweb.asm.Handle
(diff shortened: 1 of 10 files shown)


Acuadragon100:
Should also add that we need to do something about the `getExtensionInfo` calls in [CommonHooks](https://github.com/KiltMC/NeoForge/blob/kilt/1.21.1/src/main/java/net/neoforged/neoforge/common/CommonHooks.java#L1663) and [ClientHooks](https://github.com/KiltMC/NeoForge/blob/kilt/1.21.1/src/main/java/net/neoforged/neoforge/client/ClientHooks.java#L1131) because they seem to get the wrong return value.

BluSpring:
Consider: reflection

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,313 @@
...
+    private fun getCastClass(type: Type): Class<*> {
+        return when (type.sort) {
+            Type.VOID -> Void.TYPE
+            Type.BOOLEAN -> java.lang.Boolean.TYPE
+            Type.BYTE -> java.lang.Byte.TYPE
+            Type.SHORT -> java.lang.Short.TYPE
+            Type.INT -> Integer.TYPE
+            Type.LONG -> java.lang.Long.TYPE
+            Type.FLOAT -> java.lang.Float.TYPE
+            Type.DOUBLE -> java.lang.Double.TYPE
```
BluSpring: `Boolean::class.javaPrimitiveType`, etc.

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,313 @@
...
+                                val args = mutableListOf<Any?>()
+                                for (i in 0..< descriptor.argumentCount) {
+                                    args.add(proxy.getParameter(i))
+                                }
+                                args
+                            }
+                            is EnumParameters.MethodReference -> {
+                                // Luckily for us, NeoForge crashes unless these are public.
+                                // This is good because getDeclaredMethod might crash the game on servers.
+                                val method = Class.forName(parameters.owner.className).getMethod(parameters.methodName, Integer.TYPE, Class::class.java)
```
BluSpring: Is this tested on dedicated server? .... does dedicated server even work with Kilt right now, actually?

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,313 @@
...
+                                val args = mutableListOf<Any?>()
+                                for (i in 0..< descriptor.argumentCount) {
```
BluSpring: `..<` should be `until`

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,313 @@
...
+    val proxies = mutableMapOf<String, MutableMap<String, EnumParameters.FieldReference>>()
+    val indexed = mutableMapOf<String, Int>()
+
+    private fun remapPrototype(prototype: EnumPrototype): EnumPrototype {
+        return EnumPrototype(
+            prototype.owningMod,
+            KiltRemapper.remapClass(prototype.enumName, toIntermediary = !FabricLoader.getInstance().isDevelopmentEnvironment),
+            prototype.fieldName,
+            KiltRemapper.remapDescriptor(prototype.ctorDesc, toIntermediary = !FabricLoader.getInstance().isDevelopmentEnvironment),
+            KiltRemapper.remapDescriptor(prototype.fullCtorDesc, toIntermediary = !FabricLoader.getInstance().isDevelopmentEnvironment),
```
BluSpring: Does `toIntermediary` need to be specified here? By default `toIntermediary` should be redundant in prod envs, and be marked as true when the flag is specified in dev.

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,313 @@
...
+            val proxy = getEnumProxy(proxyField)
+            forceSetEnumProxyValue(proxy, enum)
+        }
+    }
+
+    private fun getEnumProxy(fieldReference: EnumParameters.FieldReference): EnumProxy<*> {
+        // Luckily for us, NeoForge crashes unless these are public.
+        // This is good because getDeclaredField might crash the game on servers.
+        return Class.forName(fieldReference.owner.className).getField(fieldReference.fieldName).get(null) as EnumProxy<*>
+    }
```
BluSpring, line 75: Is this tested on dedicated server? .... does dedicated server even work with Kilt right now, actually?

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,313 @@
...
+        }
+    }
+
+    val INITIAL_ENUM_COUNTS: Object2IntMap<String?> = Object2IntOpenHashMap<String?>()
+    const val EXTENSIBLE_ENUM: String = "net/neoforged/fml/common/asm/enumextension/IExtensibleEnum"
+
+    private fun countEnumFields(classNode: ClassNode): Int {
+        var count = 0
+        for (field in classNode.fields) {
+            if (field.desc == "L" + classNode.name + ";" && (field.access and Opcodes.ACC_STATIC) != 0) {
```
BluSpring: `"L${classNode.name};"`

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,313 @@
...
+            // I decided to set this parameter to the same value as ordinal for all values since there isn't really any good reason to allow Fabric mods to set their own values here.
+            val indexed = Annotations.getInvisible(mixinClass, IndexedEnum::class.java)
+            if (indexed != null) {
+                var index = Annotations.getValue<Number?>(indexed)
+                if (index == null) {
+                    index = 0
+                }
+				EnumExtensionLoader.indexed[targetClass.name] = index.toInt()
+				val fixedIndex = index.toInt()
+                targetClass.methods.stream().filter { method: MethodNode? -> method!!.name == "<init>" }
```
BluSpring: remove `: MethodNode?`

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,313 @@
...
+                    .forEach { constructor: MethodNode? ->
```
BluSpring: ditto

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,313 @@
...
+                        constructor!!.instructions.insertBefore(constructor.instructions.getFirst(), list)
+                    }
+            }
+        }
+    }
+
+    fun postApplyEnum(targetClass: ClassNode, mixinClass: ClassNode) {
+        if (mixinClass.interfaces.contains(EXTENSIBLE_ENUM)) {
+            val clinit =
+                targetClass.methods.stream().filter { method: MethodNode? -> method!!.name == "<clinit>" }.findFirst()
```
BluSpring: ditto

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,313 @@
...
+                        targetClass.name,
+                        extensionInfoField,
+                        extensionInfoDesc
+                    )
+                )
+                if (clinit.maxStack < 6) {
+                    clinit.maxStack = 6
+                }
+                clinit.instructions.insertBefore(clinit.instructions.getLast(), fieldSetup)
+                val getExtensionInfo = targetClass.methods.stream().filter { method: MethodNode? ->
```
BluSpring: ditto

BluSpring (changes requested):

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
Acuadragon100: It seemed to work when I tried it at least. I made a small test mod where I put all the methods for initializing the enums (including the client-only ones) in the same class file without annotating anything (because NeoForge allows that). With getDeclaredMethod/Field I got a crash because it tried to classload some client-only class, but getMethod/Field worked fine.

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
Acuadragon100: Got similar results there, though I haven't actually checked what happens when using an EnumProxy for a client-only class.

Acuadragon100: Good news, it crashes on NeoForge.

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
Acuadragon100: You're right. I tried without just now and it worked fine. Must've been me getting confused about what that parameter was actually for.

Acuadragon100:
Actually, we might have a tiny problem. It seems like the enums are inserted in a different order every time...

Yeah, using the ordinal as the id parameter might not work as well as we'd hoped... Even if we fix this in Kilt's Fabric ASM version, Fabric mods would still use the old version which will give them a scrambled order.

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+import xyz.bluspring.fork.mm.api.ClassTinkerers
+import xyz.bluspring.kilt.loader.mod.NeoForgeMod
+import xyz.bluspring.kilt.loader.remap.KiltRemapper
+import java.nio.file.Files
+import java.util.*
+import java.util.function.Consumer
+
+object EnumExtensionLoader {
+
+    val enumExtensions = mutableMapOf<String, MutableSet<EnumPrototype>>()
```
BluSpring: Would it not make more sense to make `MutableSet` a `Collection` instead, and initialize the set directly in the load method itself?

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+            prototype.fieldName,
+            KiltRemapper.remapDescriptor(prototype.ctorDesc),
+            KiltRemapper.remapDescriptor(prototype.fullCtorDesc),
+            prototype.ctorParams
+        )
+    }
+
+    fun loadEnumExtension(mod: NeoForgeMod) {
+        mod.config.getConfigList("mods").forEach { config ->
+            config.getConfigElement<String>("enumExtensions").ifPresent{ file ->
```
BluSpring: Space the curly

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+                val path = mod.owningFile.getFile().findResource(file)
+                if (Files.isRegularFile(path)) {
+                    EnumPrototype.load(mod, path).forEach{ prototype ->
```
BluSpring: ditto

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+            // I decided to set this parameter to the same value as ordinal for all values since there isn't really any good reason to allow Fabric mods to set their own values here.
+            val indexed = Annotations.getInvisible(mixinClass, IndexedEnum::class.java)
+            if (indexed != null) {
+                var index = Annotations.getValue<Number?>(indexed)
+                if (index == null) {
+                    index = 0
+                }
+				EnumExtensionLoader.indexed[targetClass.name] = index.toInt()
+				val fixedIndex = index.toInt()
+                targetClass.methods.stream().filter { method -> method!!.name == "<init>" }
```
BluSpring: stream() is not needed here, we're in Kotlin land. Remove !!, method is not nullable.

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+                    .forEach { constructor ->
+                        val list = InsnList()
+                        list.add(LabelNode())
+                        list.add(VarInsnNode(Opcodes.ILOAD, 2))
+                        list.add(VarInsnNode(Opcodes.ISTORE, 3 + fixedIndex))
+                        constructor!!.instructions.insertBefore(constructor.instructions.getFirst(), list)
```
BluSpring: Remove !!, constructor is not nullable

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+                    }
+            }
+        }
+    }
+
+    fun postApplyEnum(targetClass: ClassNode, mixinClass: ClassNode) {
+        if (mixinClass.interfaces.contains(EXTENSIBLE_ENUM)) {
+            val clinit =
+                targetClass.methods.stream().filter { method -> method!!.name == "<clinit>" }.findFirst()
+                    .orElseThrow()
```
BluSpring: `targetClass.methods.first { it.name == "<clinit>" }`

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+
+            val vanillaCount = INITIAL_ENUM_COUNTS.getOrDefault(targetClass.name, 0)
+
+            // Make sure the ordinal parameter does not change between reboots.
+            let {
+                val enumPositions: TreeMap<String, Int> = TreeMap()
```
BluSpring: `val enumPositions = TreeMap<String, Int>()`

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+                }
+                clinit.instructions.insert(clinit.instructions[setValuesIndex], fixValuesOrder)
+            }
+
+            // Fill all EnumProxy instances.
+            val proxies = proxies.get(targetClass.name)
+			proxies?.keys?.forEach(Consumer { fieldName: String? ->
+				val fillProxy = InsnList()
+				fillProxy.add(LabelNode())
+				val enumExtensionLoader = "xyz/bluspring/kilt/loader/asm/EnumExtensionLoader"
```
BluSpring: Make this a const somewhere instead

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+                )
+                if (clinit.maxStack < 6) {
+                    clinit.maxStack = 6
+                }
+                clinit.instructions.insertBefore(clinit.instructions.getLast(), fieldSetup)
+                val getExtensionInfo = targetClass.methods.stream().filter { method ->
+                    method!!.name == "getExtensionInfo" &&
+                            method.desc == "()$extensionInfoDesc" &&
+                            (method.access and Opcodes.ACC_STATIC) != 0
+                }.findFirst().orElseThrow()
```
BluSpring: `targetClass.methods.first { it.name == "getExtensionInfo" && it.desc == "()$extensionInfoDesc" && (method.access and Opcodes.ACC_STATIC) != 0 }`

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+                            usablePrototype.enumName,
+                            { TreeSet() }
+                        ).add(usablePrototype)
+                    }
+                }
+            }
+        }
+    }
+
+    private fun <T> forceSetEnumProxyValue(proxy: EnumProxy<T>, value: Any) where T: Enum<T>, T: IExtensibleEnum {
```
BluSpring: space the generic `:`s (`where T : Enum<T>, T : IExtensibleEnum`)

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+                                args
+                            }
+                        }.toTypedArray()
+                    }
+                ).build()
+            }
+        }
+    }
+
+    val INITIAL_ENUM_COUNTS: Object2IntMap<String?> = Object2IntOpenHashMap<String?>()
```
BluSpring: why is String nullable?

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -0,0 +1,393 @@
...
+    fun preApplyEnum(targetClass: ClassNode, mixinClass: ClassNode) {
+        if (mixinClass.interfaces.contains(EXTENSIBLE_ENUM)) {
+            // Note how many enum constants exist before mixin injection.
+            INITIAL_ENUM_COUNTS.put(targetClass.name, countEnumFields(targetClass))
+
+            // Handle IndexedEnums since NeoForge mods always set index to -1.
+            // I decided to set this parameter to the same value as ordinal for all values since there isn't really any good reason to allow Fabric mods to set their own values here.
+            val indexed = Annotations.getInvisible(mixinClass, IndexedEnum::class.java)
+            if (indexed != null) {
+                var index = Annotations.getValue<Number?>(indexed)
```
BluSpring: Does casting to Int not work here? Why is it set to nullable Number?

BluSpring (changes requested):

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
Acuadragon100: Ah, hmm... You mean rather than using `computeIfAbsent` I make the set in `loadEnumExtension` and set it from there? I could do that, but I'd probably just end up creating a local map in that case because it's grouped by enum class name, not by mod id. I might have misunderstood you though so feel free to clarify.

That said, I've been thinking of maybe replacing that entire map with a large mutable set instead.

Thoughts?

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
Acuadragon100: Probably because I wrote a lot of this in Java first and then let IntelliJ do the Kotlin translation.

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
Acuadragon100: I remember getting a crash while this code was still written in Java. Int seems to work fine in Kotlin though.

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
Acuadragon100: It occured to me just now that another way to solve the ordering issue might be to keep the map and make sure all NeoForge mods use the same EnumAdder for a specific enum. But I'm still a little worried about Fabric mods changing the order, even though the fact that we use a separate fork of Fabric ASM probably makes that less likely.

Issue #751, opened 2026-05-02
TFC 1.20.1 is not great


MrLogiousBanana:
### Version information

Kilt-20.1.14 + MC 1.20.1

### Expected Behavior

TerraFirmaCraft is allow player to.. Eat food and saturate? And mobs are spawn only in caves

### Actual Behavior

Food is not saturate player and mobs spawns as it was vanilla

### Reproduction Steps

1. Wait till hunger bar is not full
2. eat some food
3. wait till night or find some dark places
4. watch the mobs...

### Java Provider + Version

Java 17 (17.0.9)

### Log file

https://mclo.gs/I5u1BK5

PR #784, opened 2026-05-23
Fix player not passed to `FoodData::eat`.

Acuadragon100:
Fixes half of #751 (it mentions 2 issues). TFC overrides the forge variant of eat but ignores the vanilla one.

I wanted to keep this one separate because it introduces a new redirect which could potentially break Fabric mods with a redirect at the same place (I don't know any Fabric mods that do that though). I chose a redirect because there isn't really any good reason to call the original since the custom one will always call that one anyway (unless a mod like TFC overrides it in which case we probably don't want to run that anyway). A WrapOperation would just ignore the original, so then it would fail silently which is not ideal.


BluSpring:
This will probably need to be revisited some other time to try to avoid the redirect, but for the meantime this works well enough.

PR #788, opened 2026-05-25
Fix fluid issues with TFC

Acuadragon100:
First off, this PR fixes the bug where striders die in lava when TFC is installed. Essentially, instead of doing a check in `@ModifyReturnValue` like we did before, I change the value of bl2 (the variable returned at the end) soon after whenever it's set to true (so the fluid type has already been added to `interimCalcs`). This should hopefully keep Kilt more robust to mods that use [`CallbackInfoReturnable::setReturnValue`](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1.20.x/src/main/java/net/dries007/tfc/mixin/EntityMixin.java#L31) instead of `@ModifyReturnValue` here.

Also fixes a bug where salt and spring water don't act like water to mobs. I basically did a hack to make methods targeting the return of `updateFluidHeightAndDoFluidPushing` with a cancellable `@Inject` or a `@ModifyReturnValue`  also target a dummy method which is used to determine the cancelled return value from `kilt$useInterimValuesForCalc`. Unlike Fabric, any Forge mods targeting this method at return will always run after all the main logic.

I've already checked Valkyrien Skies, and it seems to work with this. Any other particular mods I should check?

### DIFF
--- a/src/main/java/xyz/bluspring/kilt/forgeinjects/world/entity/EntityInject.java
+++ b/src/main/java/xyz/bluspring/kilt/forgeinjects/world/entity/EntityInject.java
@@ -13,6 +13,7 @@ import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import com.llamalad7.mixinextras.sugar.Local;
 import com.llamalad7.mixinextras.sugar.Share;
+import com.llamalad7.mixinextras.sugar.ref.LocalBooleanRef;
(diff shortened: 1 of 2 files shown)


BluSpring:
Recommend checking to see if there still exists a crash with Hephaestus with this

Acuadragon100:
> Recommend checking to see if there still exists a crash with Hephaestus with this

Yeah, it doesn't seem to make any difference there unfortunately.

BluSpring:
God dammit.

PR #789, opened 2026-05-27
Fire MobSpawnEvent.FinalizeSpawn

### LINKED ISSUE #751: TFC 1.20.1 is not great

Acuadragon100:
This event is usually fired from Forge by replacing all calls to `finalizeSpawn` with a call to `ForgeEventFactory::onFinalizeSpawn` using a coremod. From what I could gather, this behaviour was disabled due to mixin conflicts.

This PR dynamically generates `@WrapMethod` mixins onto every single class overriding the `finalizeSpawn` function to call `ForgeEventFactory::kilt$onFinalizeSpawn`, which calls the original instead of always `finalizeSpawn`. I use `KiltHelper::hasMethodOverride` to only call `ForgeEventFactory::kilt$onFinalizeSpawn` once per class. I chose this approach to make sure Fabric mods also fire the event when they try to run `finalizeSpawn`.

This finally closes #751

Unfortunately, due to a [mixin in The Aether mod](https://github.com/The-Aether-Team/The-Aether/blob/1.20.1-develop/src/main/java/com/aetherteam/aether/mixin/mixins/common/ForgeEventFactoryMixin.java#L24-L29), I decided to not make `ForgeEventFactory::onFinalizeSpawn` a stub calling `ForgeEventFactory::kilt$onFinalizeSpawn` as usual. Instead, I use a mixin to wrap `finalizeSpawn` inside `ForgeEventFactory::onFinalizeSpawn` with the original from `ForgeEventFactory::kilt$onFinalizeSpawn` which is passed over with a ThreadLocal (this function is sometimes called from other threads).

The main downside with this patch is that if anyone uses `@ModifyReceiver` on `finalizeSpawn` inside `ForgeEventFactory::onFinalizeSpawn`, it will fire the FinalizeSpawn event again for the new mob. I initially tried to get around that with a ThreadLocal, but it would be tricky to make that work in cases where `finalizeSpawn` is called on other mobs inside `finalizeSpawn`. I also don't know of mods that put a `@ModifyReceiver` in that exact position. It also won't really work with Forge mods that use a `@Redirect` around `finalizeSpawn` directly inside `ForgeEventFactory::onFinalizeSpawn` and will instead always call the original set from `ForgeEventFactory::kilt$onFinalizeSpawn` which might be a Fabric redirect. There isn't really much way around that though since that's essentially equivalent to a redirect conflict. Also, we would probably have to go back to use a ThreadLocal instead of `KiltHelper::hasMethodOverride` to support running the Forge redirect.

Since this is a bit more hacky than usual and also requires a PR over at [KiltMC/NeoForge](https://github.com/KiltMC/NeoForge), I'll mark this as a draft for now.
Do you think this is reasonable, or would you prefer a different approach?


BluSpring:
Okay this is actually an insanely interesting way of doing things and I never in a million years would have thought about doing it like this, I'm gonna sit down and review this when I'm actually on a computer

Acuadragon100:
Right, one issue with this is I just found is that when Forge mods call `ForgeEventFactory::onFinalizeSpawn` directly as they should, it will run the event twice. We might need to use a ThreadLocal check instead of `hasMethodOverride`.

src/main/kotlin/xyz/bluspring/kilt/loader/asm/OverridingMethodWrappers.kt
```diff
@@ -0,0 +1,242 @@
...
+import java.nio.file.*
+import java.nio.file.attribute.BasicFileAttributes
+import java.util.*
+import java.util.function.Predicate
+import java.util.stream.Stream
+
+object OverridingMethodWrappers {
+
+    @JvmStatic
+    val mixinPackage = "xyz.bluspring.kilt.mixin"
```
BluSpring: Make this a const, capitalize to `MIXIN_PACKAGE`. If it's inaccessible from Java, use `@JvmField` instead of const.

src/main/kotlin/xyz/bluspring/kilt/loader/asm/OverridingMethodWrappers.kt
```diff
@@ -0,0 +1,242 @@
...
+
+    private const val unmappedFinalizedSpawnDesc = "(Lnet/minecraft/world/level/ServerLevelAccessor;Lnet/minecraft/world/DifficultyInstance;Lnet/minecraft/world/entity/MobSpawnType;Lnet/minecraft/world/entity/SpawnGroupData;Lnet/minecraft/nbt/CompoundTag;)Lnet/minecraft/world/entity/SpawnGroupData;"
+
+    @JvmStatic
+    val finalizeSpawnName: String = KiltRemapper.enhancedMojangRemapper?.mapMethodName("net/minecraft/world/entity/Mob", "finalizeSpawn", unmappedFinalizedSpawnDesc)!!
```
BluSpring, line 39: Why is `enhancedMojangRemapper` nullable?

src/main/kotlin/xyz/bluspring/kilt/loader/asm/OverridingMethodWrappers.kt
```diff
@@ -0,0 +1,242 @@
...
+    private val mixins = mutableListOf<String>()
+    private val classCache = mutableMapOf<Path, ClassProvider.IClassInfo>()
+
+    var enhancedMojangRemapper: KiltEnhancedRemapper? = null
+        private set
+
+    interface TargetClass {}
+
+    init {
+        enhancedMojangRemapper = KiltRemapper.enhancedMojangRemapper
```
BluSpring: Couldn't this be set directly in the field? Otherwise, the remapper could be a lateinit.

src/main/kotlin/xyz/bluspring/kilt/loader/asm/OverridingMethodWrappers.kt
```diff
@@ -0,0 +1,242 @@
...
+        // Finalize spawn exists in a bunch of places, we need a mixin where the exact value of targets depends on what classes exist at runtime.
+        val mainName = "gen/FinalizeSpawnFix"
+        val template = "xyz/bluspring/kilt/loader/asm/template/FinalizeSpawnTemplate"
+
+
+
```
BluSpring: Too many line breaks

src/main/kotlin/xyz/bluspring/kilt/loader/asm/OverridingMethodWrappers.kt
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    private fun getClassesIf(predicate: Predicate<ClassProvider.IClassInfo>): Set<String> {
+        return setOf(
+            *getClassesIf(Arrays.stream(runBlocking { KiltRemapper.getGameClassPath() }), predicate).toTypedArray(),
+            *getClassesIf(KiltHelper.getKiltPaths().stream(), predicate).toTypedArray(),
+            *getClassesIf(FabricLoader.getInstance().allMods.flatMap { container -> container.rootPaths }.stream(), predicate).toTypedArray(),
+            *getClassesIf(KiltLoader.instance.mods.stream().flatMap { it.paths.stream() }, predicate).toTypedArray()
+        )
+    }
```
BluSpring: Try to avoid streams, collections should be preferred instead

src/main/kotlin/xyz/bluspring/kilt/loader/asm/OverridingMethodWrappers.kt
```diff
@@ -0,0 +1,242 @@
...
+        mixins.add(currentName.replace("/", "."))
+    }
+
+    @JvmStatic
+    fun init() {}
+
+    @JvmStatic
+    fun getMixins(): List<String> {
+        return mixins
+    }
```
BluSpring: You could mark the mixins val as `@JvmStatic` instead.

BluSpring (changes requested):

src/main/kotlin/xyz/bluspring/kilt/loader/asm/OverridingMethodWrappers.kt
Acuadragon100: Because we want to discard it when we're done to save RAM. Only reason I store a copy of it here is because the main one is discarded too soon.

BluSpring: I might rework the remapper discard later to use some kind of reference system for freeing memory, so we can avoid things like this.

PR #791, opened 2026-05-29
Try to fix the cached size of `<clinit>` not matching the actual size of the linked list.

Acuadragon100:
For some reason the size is sometimes too small, which throws an ArrayIndexOutOfBoundsException.

### DIFF
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
@@ -22,6 +22,12 @@ object EnumExtensionLoader {
     val proxies = mutableMapOf<String, MutableMap<String, EnumParameters.FieldReference>>()
     val indexed = mutableMapOf<String, Int>()
 
+    private val insnListSizeField = InsnList::class.java.getDeclaredField("size")
+
+    init {
+        insnListSizeField.isAccessible = true
+    }
+
     private fun remapPrototype(prototype: EnumPrototype): EnumPrototype {
         return EnumPrototype(
             prototype.owningMod,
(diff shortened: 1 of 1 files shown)


src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -171,12 +171,34 @@ object EnumExtensionLoader {
...
+        while (insn != null) {
+            index++
+            if (insn == nodes.last) {
+                actualSize = index
+            }
+            insn = insn.next
+        }
+        if (actualSize != -1 && notedSize != actualSize && index == actualSize) {
+            val sizeField = nodes.javaClass.getDeclaredField("size")
+            sizeField.isAccessible = true
```
BluSpring: Cache the size field

BluSpring (changes requested):

src/main/kotlin/xyz/bluspring/kilt/loader/asm/EnumExtensionLoader.kt
```diff
@@ -22,6 +22,12 @@ object EnumExtensionLoader {
     val proxies = mutableMapOf<String, MutableMap<String, EnumParameters.FieldReference>>()
     val indexed = mutableMapOf<String, Int>()
 
+    private val insnListSizeField = InsnList::class.java.getDeclaredField("size")
+
+    init {
+        insnListSizeField.isAccessible = true
+    }
```
BluSpring: This can be moved into a `.apply` after the getDeclaredField

```kt
InsnList::class.java.getDeclaredField("size").apply {
  isAccessible = true
}
```

BluSpring (changes requested):

PR #794, opened 2026-05-30
Make RetargetingLocalModifier match target even if wrapped in LocalRef.

Acuadragon100:
Possible to disable on a case-by-case basis if necessary.

I use the signature to unbox LocalRef, it worked on a small dummy mod I made at least.

### DIFF
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/mixin/modifications/LocalPair.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/mixin/modifications/LocalPair.kt
@@ -5,7 +5,8 @@ import org.objectweb.asm.Type
 
 data class LocalPair(
     val descriptor: String,
-    val local: Local
+    val local: Local,
+    val unboxRef: Boolean = true
 ) {
     constructor(type: Type, local: Local) : this(type.descriptor, local)
 }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/xyz/bluspring/kilt/loader/mixin/modifications/modifiers/RetargetingLocalModifier.kt
```diff
@@ -65,7 +140,8 @@ data class RetargetingLocalModifier(
                     val argsOnly = values["argsOnly"]
                     val type = values["type"]
 
-                    val mapping = paramToLocalMapping.filter { it.key.descriptor == descriptor.descriptor
+                    val mapping = paramToLocalMapping.filter {
+                            it.key.descriptor == (if (it.key.unboxRef) unboxRefDescriptor(descriptor, signature).descriptor else descriptor.descriptor)
```
BluSpring: Try doing `(if (it.key.unboxRef) unboxRefDescriptor(descriptor, signature) else descriptor).descriptor`

src/main/kotlin/xyz/bluspring/kilt/loader/mixin/modifications/LocalPair.kt
```diff
@@ -5,7 +5,8 @@ import org.objectweb.asm.Type
 
 data class LocalPair(
```
BluSpring, line 7: Do this change for me rq, make this an `@JvmRecord`

BluSpring (changes requested):

PR #795, opened 2026-05-30
Fix mixin issues with No Man's Land

Acuadragon100:
Depends on #794

Decided to do a simple mod-specific fix for the named local for now.

### DIFF
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/mixin/modifications/KiltMixinModifications.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/mixin/modifications/KiltMixinModifications.kt
@@ -51,6 +51,15 @@ object KiltMixinModifications {
             mapOf(
                 LocalPair("Lnet/minecraft/core/Direction;", Local(name = arrayOf("direction"))) to Local(ordinal = 0)
             )
+        ),
+
+        // Fix No Man's Land ServerPlayerMixin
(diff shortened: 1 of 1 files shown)


src/main/kotlin/xyz/bluspring/kilt/loader/mixin/modifications/KiltMixinModifications.kt
```diff
@@ -51,6 +54,15 @@ object KiltMixinModifications {
...
                 LocalPair("Lnet/minecraft/core/Direction;", Local(name = arrayOf("direction"))) to Local(ordinal = 0)
             )
+        ),
+
+        // Fix No Man's Land ServerPlayerMixin
+        // https://github.com/Alchemists-Of-Yore/No-Mans-Land/blob/1.21.1/src/main/java/com/farcr/nomansland/common/mixin/ServerPlayerMixin.java#L35-L41
+        // This lambda is added by NeoForge and contains some of the logic. Luckily for us, they have the same signature.
+        NameRemappingAnnotationModifier(
+            owner = "net/minecraft/server/level/ServerPlayer",
+            methods = listOf($$"lambda$startSleepInBed$13"),
```
BluSpring: Also add the descriptor-based lambda target, some mods may target that too.

BluSpring (changes requested):

PR #796, opened 2026-05-30
Remap this$0 in dev as well.

Acuadragon100:


src/main/kotlin/xyz/bluspring/kilt/loader/remap/KiltEnhancedRemapper.kt
```diff
@@ -59,7 +59,7 @@ class KiltEnhancedRemapper(private val provider: ClassProvider, private val file
     }
 
     override fun mapFieldName(owner: String, name: String, descriptor: String): String {
-        if (FabricLoader.getInstance().isDevelopmentEnvironment && !KiltRemapper.forceProductionRemap) {
+        if (FabricLoader.getInstance().isDevelopmentEnvironment && !KiltRemapper.forceProductionRemap && name != "this$0") {
```
BluSpring: Target all numeric `this` calls, it's possible for nested inner classes to have multiple of these, even if it's very unlikely.

BluSpring (changes requested):

PR #797, opened 2026-05-30
Added LivingEntity bed injects.

Acuadragon100:
This PR makes Kilt run `isBed`, `getBedDirection` and `setBedOccupied` checks for NeoForge blocks, as well as fire `CanContinueSleepingEvent` from LivingEntity.

This fixes a bug where straw beds in No Man's Land don't work.

Everything is handled via Fabric API, with the exception of CanContinueSleepingEvent which I added via mixin since I didn't feel that Fabric API had the specific event for it.

Let me know if you know any Fabric mods or other NeoForge mods that add custom beds to test this with.
I have not checked if `setBedOccupied` actually works because I don't know any mods for 1.21.1 that use it which also work at the moment.

### DIFF
--- a/src/main/java/xyz/bluspring/kilt/injects/world/entity/LivingEntityInject.java
+++ b/src/main/java/xyz/bluspring/kilt/injects/world/entity/LivingEntityInject.java
@@ -17,7 +17,6 @@ import com.llamalad7.mixinextras.sugar.ref.LocalDoubleRef;
 import com.llamalad7.mixinextras.sugar.ref.LocalFloatRef;
 import com.llamalad7.mixinextras.sugar.ref.LocalRef;
 import net.minecraft.core.BlockPos;
-import net.minecraft.core.Direction;
 import net.minecraft.core.Holder;
 import net.minecraft.core.component.DataComponentType;
 import net.minecraft.core.particles.ParticleOptions;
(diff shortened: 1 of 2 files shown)


BluSpring:
Try Comforts out

src/main/kotlin/xyz/bluspring/kilt/Kilt.kt
```diff
@@ -106,6 +115,57 @@ class Kilt : ModInitializer {
...
+        }
+
+        EntitySleepEvents.MODIFY_SLEEPING_DIRECTION.register { entity, pos, direction ->
+            val state: BlockState = entity.level().getBlockState(pos)
+            if (
+                (
+                    hasMethodOverride(
+                        state.block.javaClass, IBlockExtension::class.java, "getBedDirection",
+                        BlockState::class.java, LevelReader::class.java, BlockPos::class.java
+                    ) || // If bed is not BebBlock we need to run the NeoForge getBedDirection for the correct result even if not overriden.
```
BluSpring: typo

src/main/kotlin/xyz/bluspring/kilt/Kilt.kt
```diff
@@ -29,6 +37,7 @@ import xyz.bluspring.kilt.client.KiltClient
 import xyz.bluspring.kilt.loader.KiltLoader
 import xyz.bluspring.kilt.mixin.MinecraftServerAccessor
 import xyz.bluspring.kilt.util.KiltHelper
+import xyz.bluspring.kilt.util.KiltHelper.hasMethodOverride
```
BluSpring: Don't shortcut hasMethodOverride

BluSpring (changes requested):

Acuadragon100:
> Try Comforts out

Both the NeoForge and Fabric versions seem to work fine.

src/main/java/xyz/bluspring/kilt/injects/world/entity/LivingEntityInject.java
```diff
@@ -658,7 +645,13 @@ public boolean removeEffectsCuredBy(EffectCure cure) {
...
     }
 
-    // TODO: bed checks
+    // The rest of the bed checks are implemented via Fabric API.
+
+    @WrapMethod(method = "checkBedExists")
+    private boolean kilt$checkBedExists(Operation<Boolean> original) {
+        boolean hasBed = original.call();
+        return EventHooks.canEntityContinueSleeping((LivingEntity) (Object) this, hasBed ? null : Player.BedSleepingProblem.NOT_POSSIBLE_NOW);
+    }
```
BluSpring: Is this unable to be a ModifyReturnValue?

BluSpring (changes requested):

src/main/java/xyz/bluspring/kilt/injects/world/entity/LivingEntityInject.java
Acuadragon100: Would probably work fine. Probably just me still thinking about the fluid issues where everyone used `CallbackInfoReturnable::setReturnValue` instead of `@ModifyReturnValue` and it caused issues.

PR #798, opened 2026-05-31
Fix `EntityJoinLevelEvent` never being fired for non-player entities on the server side (and being fired twice for players).

Acuadragon100:
Fixes the mob variant texture issues in No Man's Land.

### DIFF
--- a/src/main/java/xyz/bluspring/kilt/injects/world/level/entity/PersistentEntitySectionManagerInject.java
+++ b/src/main/java/xyz/bluspring/kilt/injects/world/level/entity/PersistentEntitySectionManagerInject.java
@@ -38,7 +38,7 @@ public abstract class PersistentEntitySectionManagerInject<T extends EntityAcces
 
     @Inject(method = "addEntity", at = @At("HEAD"), cancellable = true)
     private void kilt$callEntityJoinLevelEvent(T entity, boolean worldGenSpawned, CallbackInfoReturnable<Boolean> cir) {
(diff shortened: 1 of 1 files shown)


src/main/java/xyz/bluspring/kilt/injects/world/level/entity/PersistentEntitySectionManagerInject.java
```diff
@@ -38,7 +38,7 @@ public boolean addNewEntityWithoutEvent(T entity) {
 
     @Inject(method = "addEntity", at = @At("HEAD"), cancellable = true)
     private void kilt$callEntityJoinLevelEvent(T entity, boolean worldGenSpawned, CallbackInfoReturnable<Boolean> cir) {
-        if (this.kilt$callWithoutEvent.getAndSet(false) && entity instanceof Entity e && NeoForge.EVENT_BUS.post(new EntityJoinLevelEvent(e, e.level(), worldGenSpawned)).isCanceled())
+        if (!this.kilt$callWithoutEvent.getAndSet(false) && entity instanceof Entity e && NeoForge.EVENT_BUS.post(new EntityJoinLevelEvent(e, e.level(), worldGenSpawned)).isCanceled())
```
BluSpring, line 42: could you add a comment here that tells future me to not remove the exclamation because i'm pretty sure i did that myself thinking "oh wow this probably shouldn't be inverted"

BluSpring (changes requested):

PR #799, opened 2026-05-31
Fix some options being reloaded on startup in 1.21.1

Acuadragon100:
This fixes the gui scale reset issue which reappeared on 1.21.1.

Basically, some options are processed in a separate method called `processDumpedOptions`, which I just wrap in a `@WrapWithCondition`.
There is also one extra setting under `onboardAccessibility` which I thought would be a better target to stop the limited rules.

The `@Coerce` is there because `Options.OptionAccess` is package-private, and I didn't feel like adding an access widener for it when we don't need to use it.


BluSpring:
Huh, didn't know Coerce did that. Good to know for next time.

PR #800, opened 2026-06-01
Give NeoForge mods a module layer and access to scan data.

Acuadragon100:
Also needed by Twilight Forest beanification.

Marking as draft since it also bumps the fml module.


BluSpring:
Please tell me this doesn't make Neo mods into modules, I've been trying to ignore that aspect for so long O/-<

Acuadragon100:
I don't think it does at least. This just sets up a single module layer so Twilight Forest can access it. Are there any particular mods that would break completely with modules enabled? Because Aeronautics still seems to work.

BluSpring:
Okay, actually looking at it now while being awake it's not as bad as I thought.

PR #804, opened 2026-06-03
CoreMod Remapping

Acuadragon100:
This is my initial attempt at remapping CoreMods.
Basically, I'm just remapping the class to Mojmap, running it through all the transformers, then remapping it back to intermediary. To do this we need an inverse enhanced remapper which I store in a `TemporaryStartupContainer` so it can be deleted when no longer needed. Although at the moment it's never deleted because some classes that have transformers applied (primarily RaidCommand) never get classloaded.

This depends on https://github.com/KiltMC/Fabric-ASM/pull/2, otherwise some MixinExtras modifiers won't apply properly after the class has been remapped back and forth.


Acuadragon100:
I wonder how we should deal with classes that never get classloaded, because client-only classes will prevent the mappings from getting discarded on servers. I guess we could just classload them all manually and catch any errors that occur, but I'm a little worried about mods that rely on classes being classloaded at a certain point.

src/main/kotlin/xyz/bluspring/kilt/loader/remap/KiltRemapper.kt
```diff
@@ -289,41 +291,49 @@ object KiltRemapper {
...
+                        // add mapped path too
+                        *runBlocking { getGameClassPath() },
+                        // Add all NeoForge mods to the library path, because dependencies don't have to be specified
+                        // in order to use mods lmao
+                        *modLoadingQueue.map { mod -> mod.path }.toTypedArray()
+                    ).forEach {
+                        addLibrary(it)
+                    }
+                }.build()
+            }
```
BluSpring: I thought I did this already, but just copy over the changes made in 1.20.1 for creating multiple remappers (#709). imo it looks cleaner.

src/main/kotlin/xyz/bluspring/kilt/util/TemporaryStartupContainer.kt
BluSpring, line 1: To be honest, I think this is a better fit for 1.20.1. Because of the volatility of 1.21.1 due to Java coremods, I feel like it's necessary at this point to just eat the memory usage. This can be revisited some other time.

BluSpring (changes requested):

PR #805, opened 2026-06-03
Inject mods with CoreMods before loading CoreMods.

Acuadragon100:


src/main/kotlin/xyz/bluspring/kilt/loader/asm/coremod/CoreModLoader.kt
```diff
@@ -122,6 +124,16 @@ object CoreModLoader {
...
             return
+
+        for (mod in KiltLoader.instance.mods) {
+            if (mod.modFile == null) continue
+            val coreModStream = mod.getFile("META-INF/services/net.neoforged.neoforgespi.coremod.ICoreMod")
+            if (coreModStream != null) {
+                coreModStream.close()
+                KnitLoader.instance.injectIntoClasspath(mod.modFile.toPath())
+            }
+        }
```
BluSpring, line 135: can't we just.. move the Java coremod loading to after when mods actually get injected?

BluSpring (changes requested):

src/main/kotlin/xyz/bluspring/kilt/loader/asm/coremod/CoreModLoader.kt
BluSpring: wait, no we can't, that's too late in the process isn't it

Acuadragon100: I haven't tried it, but I suspected something might go wrong when doing it later. This appears to work.

BluSpring: I'm not a big fan of how this is going right now, this is a weird special case. I'll look into it myself.

PR #806, opened 2026-06-03
Shadow language instead of adding a new field.

Acuadragon100:


src/main/java/xyz/bluspring/kilt/injects/server/level/ServerPlayerInject.java
```diff
@@ -256,7 +256,7 @@ public OptionalInt openMenu(MenuProvider menuProvider, Consumer<RegistryFriendly
             ci.cancel();
     }*/
 
-    @Unique private String language = "en_us";
+    @Shadow private String language = "en_us";
```
BluSpring: Probably best to remove the `= "en_us"`

BluSpring (changes requested):

PR #809, opened 2026-06-04
Port method remapping fix from 1.20.1

Acuadragon100:
Ports the remapper fix from #748

Fixes [Blueprint's PlayerRendererMixin](https://github.com/team-abnormals/blueprint/blob/01258fce8f85d9eadc5455436287eb07b5e0a7c3/src/main/java/com/teamabnormals/blueprint/core/mixin/client/PlayerRendererMixin.java#L20) being remapped incorrectly in production.


BluSpring:
i forgot we already solved this problem

PR #813, opened 2026-06-05
Make sure mods can access their own mod container while classloading.

Acuadragon100:
Needed to prevent crash on startup with Twilight Forest since it tries to access its own mod container during `<clinit>`.

I couldn't figure out how to change the getter for `container` directly without removing the lateinit parameter, so I just renamed the field instead.


BluSpring:
Fixed this in an alternative way, see 0b2390f891cb0c3bbc7663b81cc068861fa04f6d

PR #817, opened 2026-06-05
Fix Blueprint's PlayerRendererMixin, and support renaming to multiple targets.

Acuadragon100:
A follow-up to #809 to actually fix the Blueprint bug, as well as #810 to fix global remappers not being remapped during registration.


Acuadragon100:
In hindsight, a potential alternative solution could be to change `MixinRemapper` to remap methods to multiple target methods when we have ambiguous method targets like this.

PR #823, opened 2026-06-07
Clear and re-bake on registry sync.

Acuadragon100:
Follow-up to https://github.com/KiltMC/NeoForge/pull/15 to make sure `onClear` is called for NeoForge and vanilla registries.


Acuadragon100:
I changed it so this no longer depends on https://github.com/KiltMC/NeoForge/pull/15.
Just note that we still need https://github.com/KiltMC/NeoForge/pull/15 or something similar to it since `Block::BLOCK_STATE_REGISTRY` needs to be cleared after all entries have been added but before freezing on servers.

PR #833, opened 2026-06-10
Created ModifiedCloneWorkaroundLoader to handle signature-changing patches.

Acuadragon100:
This PR resolves the crash when NeoForge mods use RangedBowAttackGoal or RangedCrossbowAttackGoal with non-monster entities.

Basically, it copies the RangedBowAttackGoal and RangedCrossbowAttackGoal after they have been modified by mixins and coremods and remaps the NeoForge mods to use that instead. I also inject the fallback Monster constructor manually to the NeoForge clone since it would just cause issues for Fabric mods.

Note that while this should work with RangedCrossbowAttackGoal without any changes, we will need to clean up RangedBowAttackGoal a bit. One way to do that would be to revert the following commits:
- 2fbfb20a1e076bff336776b492bdb1c7c9ce43be
- b72ed9306591f1e9e7617da7615a5979fb438c12

I was thinking of doing that in a separate PR.

Would you like to keep the changes to [KiltEarlyRiser](https://github.com/KiltMC/Kilt/commit/2fbfb20a1e076bff336776b492bdb1c7c9ce43be#diff-05eb33b4653f494353e9b5e2b9b71cc889aa008a99277468ecb0b98fab636f97)?

### DIFF
--- a/src/main/kotlin/xyz/bluspring/kilt/util/ModifiedCloneWorkaroundLoader.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/util/ModifiedCloneWorkaroundLoader.kt
@@ -1,0 +1,305 @@
+package xyz.bluspring.kilt.util
+
+import org.objectweb.asm.ClassWriter
+import org.objectweb.asm.Handle
+import org.objectweb.asm.Opcodes
+import org.objectweb.asm.Type
+import org.objectweb.asm.tree.*
+import xyz.bluspring.fork.mm.api.ClassTinkerers
+import xyz.bluspring.kilt.loader.remap.KiltRemapper
+
+object ModifiedCloneWorkaroundLoader {
+
+    fun init() {
+        val monsterType = KiltRemapper.remapClass("net/minecraft/world/entity/monster/Monster")
+        val monsterDesc = "L$monsterType;"
+        val mobType = KiltRemapper.remapClass("net/minecraft/world/entity/Mob")
+        val mobDesc = "L$mobType;"
+        val rangedAttackMob = KiltRemapper.remapDescriptor("Lnet/minecraft/world/entity/monster/RangedAttackMob;")
+        run {
+            val bowAttackGoal = "net/minecraft/world/entity/ai/goal/RangedBowAttackGoal"
+            val bowAttackGoalClone = "xyz/bluspring/kilt/workarounds/RangedBowAttackGoalWorkaround"
+            addTransformedClone(
+                KiltRemapper.remapClass(bowAttackGoal),
+                bowAttackGoalClone,
+            ) { classNode ->
+                replaceType(classNode, monsterType, mobType)
+
+                val fallbackConstructor = MethodNode(
+                    Opcodes.ACC_PUBLIC, "<init>", "(${monsterDesc}DIF)V",
+                    "<M:$monsterDesc:$rangedAttackMob>(TM;DIF)V", null
+                ).apply {
+                    val startLabel = LabelNode()
+                    this.instructions.add(startLabel)
+                    this.instructions.add(VarInsnNode(Opcodes.ALOAD, 0))
+                    this.instructions.add(VarInsnNode(Opcodes.ALOAD, 1))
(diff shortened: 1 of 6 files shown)


BluSpring:
Might as well clean it up in the same PR. I think the early riser changes can be removed if this exists.

src/main/kotlin/xyz/bluspring/kilt/util/ModifiedCloneWorkaroundLoader.kt
```diff
@@ -0,0 +1,317 @@
+package xyz.bluspring.kilt.util
```
BluSpring: Move to workarounds instead

src/main/kotlin/xyz/bluspring/kilt/util/ModifiedCloneWorkaroundLoader.kt
```diff
@@ -0,0 +1,317 @@
...
+import org.objectweb.asm.Opcodes
+import org.objectweb.asm.Type
+import org.objectweb.asm.tree.*
+import xyz.bluspring.fork.mm.api.ClassTinkerers
+import xyz.bluspring.kilt.loader.remap.KiltRemapper
+
+object ModifiedCloneWorkaroundLoader {
+
+    val REMAPPED_TYPES = mutableMapOf<String, String>()
+    val TRANSFORMERS = mutableListOf<() -> Unit>()
```
BluSpring: Just use `mutableListOf<Runnable>()`

src/main/kotlin/xyz/bluspring/kilt/util/ModifiedCloneWorkaroundLoader.kt
```diff
@@ -0,0 +1,317 @@
...
+                            "speedModifier", "D", null,
+                            startLabel, endLabel, 2
+                        )
+                    )
+                    this.localVariables.add(
+                        LocalVariableNode(
+                            "attackRadius", "F", null,
+                            startLabel, endLabel, 4
+                        )
+                    )
```
BluSpring: Inline the LVs

src/main/kotlin/xyz/bluspring/kilt/util/ModifiedCloneWorkaroundLoader.kt
```diff
@@ -0,0 +1,317 @@
...
+                            "attackIntervalMin", "I", null,
+                            startLabel, endLabel, 4
+                        )
+                    )
+                    this.localVariables.add(
+                        LocalVariableNode(
+                            "attackRadius", "F", null,
+                            startLabel, endLabel, 5
+                        )
+                    )
```
BluSpring: Inline the LVs

src/main/kotlin/xyz/bluspring/kilt/util/ModifiedCloneWorkaroundLoader.kt
```diff
@@ -0,0 +1,317 @@
...
+            }
+            val cw = ClassWriter(0)
+            cw.visit(
+                52,
+                Opcodes.ACC_PUBLIC,
+                cloneClass,
+                null,
+                "java/lang/Object",
+                null
+            )
```
BluSpring: Inline, but you might want to document what the magic constant `52` is for.

src/main/kotlin/xyz/bluspring/kilt/util/ModifiedCloneWorkaroundLoader.kt
```diff
@@ -0,0 +1,317 @@
...
+            }
+        }
+        return type
+    }
+
+    private fun mapHandle(handle: Handle, oldTypeName: String, newTypeName: String): Handle {
+        if (handle.owner == oldTypeName) {
+            return Handle(
+                handle.tag, newTypeName, handle.name, handle.desc, handle.isInterface
+            )
```
BluSpring: Inline

BluSpring (changes requested):

PR #856, opened 2026-06-16
Backport Relevant Server Injects from 1.21.1 + Added MarkDirtyMap

Acuadragon100:
Essentially just a backport of 3927e784d24409b66751d307b9ed1d8fa106a59c to 1.20.1.

But I noticed that the [Tardis Mod](https://modrinth.com/mod/tardis-mod) would crash with a ConcurrentModificationException when an interior dimension was created, so I made Kilt always use the world array. To make sure Fabric mods update the world array properly, we now have a MarkDirtyMap which wraps basically every single edge case where the map is modified that I could think of and runs the mark dirty function. This technically makes the Forge markDirty function redundant though. What I do know is that it works with the Fabric version of [TARDIS Refined](https://modrinth.com/mod/tardis-refined), which also creates dimensions dynamically.

I wrote MarkDirtyMap in Java because it seemed easier considering we are wrapping a regular Java map.

### DIFF
--- a/src/main/java/xyz/bluspring/kilt/forgeinjects/server/MinecraftServerInject.java
+++ b/src/main/java/xyz/bluspring/kilt/forgeinjects/server/MinecraftServerInject.java
@@ -1,15 +1,12 @@
 // TRACKED HASH: 0a24b5b7aafa1b5d82426467010c877512d3e484
 package xyz.bluspring.kilt.forgeinjects.server;
 
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Map;
+import java.util.*;
 import java.util.function.BooleanSupplier;
 
 import com.google.common.collect.Maps;
 import com.google.gson.Gson;
 import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
-import com.llamalad7.mixinextras.injector.ModifyReceiver;
 import com.llamalad7.mixinextras.injector.ModifyReturnValue;
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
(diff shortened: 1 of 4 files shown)


src/main/java/xyz/bluspring/kilt/helpers/MarkDirtyMap.java
```diff
@@ -0,0 +1,494 @@
...
+            fixArray(array);
+            return array;
+        }
+
+        @Override
+        public boolean removeIf(@NotNull Predicate<? super Entry<K, V>> filter) {
+            return super.removeIf(e -> filter.test(new MarkDirtyEntry(e)));
+        }
+
+        public class MarkDirtyEntrySetSpliterator implements Spliterator<Entry<K, V>> {
```
BluSpring: Does this have a use being public? Otherwise this should probably be private static

src/main/java/xyz/bluspring/kilt/forgeinjects/server/MinecraftServerInject.java
```diff
@@ -63,5 +96,86 @@ public String getStatusJson() {
...
+    // Kilt TODO: do we need pack stuff?
+
+    @Unique private Map<ResourceKey<Level>, long[]> perWorldTickTimes = Maps.newIdentityHashMap();
+
+    @Override
+    public long[] getTickTime(ResourceKey<Level> dim) {
+        return perWorldTickTimes.get(dim);
+    }
+
+    public synchronized Map<ResourceKey<Level>, ServerLevel> forgeGetWorldMap() {
```
BluSpring, line 154: `@Override`

src/main/java/xyz/bluspring/kilt/forgeinjects/server/MinecraftServerInject.java
```diff
@@ -1,57 +1,90 @@
...
-    }
-
-    @ModifyReturnValue(method = "getServerModName", at = @At("RETURN"), remap = false)
-    private String kilt$appendKiltToServerBranding(String original) {
-        return original + " + kilt";
-    }
+    @Shadow
+    @Final
+    @Mutable
+    private Map<ResourceKey<Level>, ServerLevel> levels;
```
BluSpring: Inline the annotations

src/main/java/xyz/bluspring/kilt/forgeinjects/server/MinecraftServerInject.java
```diff
@@ -1,57 +1,90 @@
...
+
     @ModifyExpressionValue(method = "tickServer", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/MinecraftServer;buildServerStatus()Lnet/minecraft/network/protocol/status/ServerStatus;"))
     private ServerStatus kilt$resetStatusCache(ServerStatus original) {
         this.resetStatusCache(original);
         return original;
     }
 
+    @Unique private static final Gson GSON = new Gson();
+    @Unique private String cachedServerStatus;
+    @Unique private void resetStatusCache(ServerStatus status) {
```
BluSpring: This one should be a newlined annotation

BluSpring (changes requested):

src/main/java/xyz/bluspring/kilt/helpers/MarkDirtyMap.java
Acuadragon100: Private is fine I think, especially since it's a non-static inner class at the moment. But to make it static I would probably also need to make `MarkDirtyEntry` static as well, or do a bunch of ugly class casts. Maybe more of the inner classes should be static?

BluSpring: Didn't realize there were more inner classes, yeah probably.

PR #859, opened 2026-06-17
Replace the `attributeslib_potion_gui_tooltips` CoreMod with a Mixin.

Acuadragon100:
Re-implements the following CoreMod as a mixin:
https://github.com/Shadows-of-Fire/Apothic-Attributes/blob/1.20/src/main/resources/coremods/potion_gui_tooltips.js

We need to disable it completely since it breaks the bytecode. I decided to use ServiceLoader for this so we don't have to worry about registering the event too late.

### DIFF
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/asm/coremod/CoreModLoader.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/asm/coremod/CoreModLoader.kt
@@ -2,8 +2,10 @@ package xyz.bluspring.kilt.loader.asm.coremod
 
 import com.google.gson.JsonParser
 import net.fabricmc.loader.impl.gui.FabricGuiEntry
+import xyz.bluspring.kilt.event.LoadCoreModEvent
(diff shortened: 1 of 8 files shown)


src/main/kotlin/xyz/bluspring/kilt/loader/asm/coremod/CoreModLoader.kt
```diff
@@ -67,6 +69,12 @@ object CoreModLoader {
     val loadedCoreMods = mutableListOf<CoreMod>()
     val enableCoreMods = KiltFlags.DISABLE_COREMODS
 
+    val loadEvent = LoadCoreModEvent {
+        coremod: CoreMod -> ServiceLoader.load(LoadCoreModEvent::class.java).toList().all { event ->
+            event.loadCoreMod(coremod)
+        }
+    }
```
BluSpring, line 76: this is the first time i've seen an event done like this and quite frankly i'm kinda jealous i never thought of this

PR #860, opened 2026-06-17
Add bridge for Jade

### LINKED ISSUE #848: jade bridge
### Version information

1.21.1

### What you would like added

Jade addons bridge would be cool

### Additional information

Jade addons on both platforms implement exactly the same APIs; they're all marked with `@snownee.jade.api.WailaPlugin`, except on Fabric it needs to be added as an entrypoint of type `jade`.

vgskye:
Closes #848

### DIFF
--- a/compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/jade/CommonProxyMixin.java
+++ b/compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/jade/CommonProxyMixin.java
@@ -1,0 +1,68 @@
+package xyz.bluspring.kilt.compat.fabric.mixin.jade;
+
+import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
+import com.moulberry.mixinconstraints.annotations.IfModLoaded;
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.ModContainer;
+import net.fabricmc.loader.api.entrypoint.EntrypointContainer;
+import net.neoforged.fml.ModList;
+import net.neoforged.neoforgespi.language.ModFileScanData;
+import org.objectweb.asm.Type;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Pseudo;
+import org.spongepowered.asm.mixin.Unique;
+import org.spongepowered.asm.mixin.injection.At;
+import snownee.jade.api.IWailaPlugin;
+import snownee.jade.api.WailaPlugin;
+import snownee.jade.util.CommonProxy;
+import xyz.bluspring.kilt.Kilt;
+import xyz.bluspring.kilt.loader.KiltLoader;
+import xyz.bluspring.kilt.loader.mod.NeoForgeMod;
+
+import java.util.List;
+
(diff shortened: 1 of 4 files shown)


compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/jade/CommonProxyMixin.java
```diff
@@ -0,0 +1,68 @@
...
+                                    return fabricMod;
+                                }
+
+                                @Override
+                                public String getDefinition() {
+                                    return className + " (kilt)";
+                                }
+                            });
+                        } catch (Exception e) {
+                            Kilt.Companion.getLogger().error("Failed to register Forge Jade entrypoint {} for mod {} ({})!", annotation.clazz().getClassName(), mod.getDisplayName(), mod.getModId(), e);
```
BluSpring: Should probably be named NeoForge instead here in this case.

compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/jade/CommonProxyMixin.java
```diff
@@ -0,0 +1,68 @@
...
+    @Unique
+    private static final Type kilt$pluginAnnotationType = Type.getType(WailaPlugin.class);
+
+    @ModifyExpressionValue(method = "loadComplete", at = @At(value = "INVOKE", target = "Lnet/fabricmc/loader/api/FabricLoader;getEntrypointContainers(Ljava/lang/String;Ljava/lang/Class;)Ljava/util/List;"))
+    private static List<EntrypointContainer<IWailaPlugin>> kilt$jade$loadForgeJadePlugins(List<EntrypointContainer<IWailaPlugin>> original) {
+        for (NeoForgeMod mod : KiltLoader.Companion.getInstance().getMods()) {
+            for (ModFileScanData.AnnotationData annotation : mod.getScanData().getAnnotations()) {
+                if (annotation.annotationType().equals(kilt$pluginAnnotationType)) {
+                    String modid = ((String) annotation.annotationData().get("value"));
+                    if (modid == null || ModList.get().isLoaded(modid)) {
```
BluSpring: `KiltLoader#hasMod(modid)` should be faster and safer to call into - `ModList#isLoaded` usually does a scan for Fabric mods too, while `KiltLoader#hasMod` is directly a search in Kilt's mods list (and is usually what `ModList#isLoaded` defers to anyway)

BluSpring (changes requested):

compat/fabric-compats/src/main/java/xyz/bluspring/kilt/compat/fabric/mixin/jade/CommonProxyMixin.java
vgskye: I made this call into `ModList#isLoaded` just in case some neoforge jade addon was predicating on the existence of a multiloader mod we only have the fabric copy of, but none of ModList's name changer logic is necessary because Jade does a FabricLoader isModLoaded check right after this and I don't think any mod actually uses this predicate condition. I'll probably simplify to a FabricLoader check instead

PR #863, opened 2026-06-17
Improve CC:T bridging

vgskye:
the original pr got merged while I was writing this so second PR it is


BluSpring:
Should probably mark as draft while in the process of writing, I normally wait for changes to be made after requesting 'em and after that merge time

vgskye:
yeah, will do that for future PRs

PR #867, opened 2026-06-19
Fix Quark

vgskye:
Quark now works as long as you don't have JEI. the JEI plugin crashes due to an upstream Quark bug that Kilt worsens; see VazkiiMods/Quark#5596


src/main/kotlin/xyz/bluspring/kilt/loader/mixin/KiltAnnotationAdjuster.kt
```diff
@@ -20,9 +20,9 @@ class KiltAnnotationAdjuster : MixinAnnotationAdjuster {
     ): AdjustableAnnotationNode? {
         if (mixinClassName == "org.violetmoon.quark.mixin.mixins.client.ContainerScreenMixin" && annotationNode.`is`(ModifyVariable::class.java)) {
             val variableNode = annotationNode.`as`(AdjustableModifyVariableNode::class.java)
-            if (variableNode.method[0].equals("mouseClicked(DDI)Z"))
+            if (variableNode.method[0].equals("method_25402(DDI)Z") || variableNode.method[0].equals("mouseClicked(DDI)Z"))
                 return variableNode.withIndex({ 14 })
-            else if (variableNode.method[0].equals("mouseReleased(DDI)Z"))
+            else if (variableNode.method[0].equals("method_25406(DDI)Z") || variableNode.method[0].equals("mouseReleased(DDI)Z"))
```
BluSpring: Thinking about it, could these not use the remapper? It should still be active by this point.

BluSpring (changes requested):

src/main/kotlin/xyz/bluspring/kilt/loader/mixin/KiltAnnotationAdjuster.kt
vgskye: I think MixinAnnotationAdjuster runs whenever? does the remapper stay active forever?

BluSpring: It should, because it's kinda required to.

Issue #878, opened 2026-06-21
Crafting consumes ingredients that are not supposed to be consumed in ProjectE


Bass253:
### Version information

20.1.17

### Expected Behavior

Items like philosopher's stone, Volcanite Amulet and Evertide Amulet should stay in crafting table after crafting. 

### Actual Behavior

Philosopher's stone, Volcanite Amulet and Evertide Amulet are consumed after crafting.

### Reproduction Steps

1. Install kilt and projecte.
2. Create a new world.
3. Craft something with philosopher's stone.
4. Philosopher's stone disappears.

### Java Provider + Version

Oracle 17.0.12

### Log file

https://mclo.gs/tHlDa54

Acuadragon100:
Fixed in #879

PR #879, opened 2026-06-21
Fix forge crafting remainder items not being used.

Acuadragon100:
Hooks into Fabric API so `FabricItem::getRecipeRemainder` attempts to use the forge crafting remainder if overridden.

I considered adding manual injects into BannerDuplicateRecipe and BookCloningRecipe as well since Fabric API doesn't handle them, but wasn't sure if any forge mods would actually use them.

 I also added a stack overflow mitigation to the forge module, let me know if you'd like me to move that into a mixin or something.


Acuadragon100:
Closes #878

BluSpring:
Honestly yeah the mixin approach would be better, so it can be pretty easily copied into 1.21.1 too. It's definitely not needed for 26.1 though once that rolls around.

PR #889, opened 2026-06-24
Fix Ars Nouveau failing to remap due to a malformed jar

vgskye:
hey ars, what. how.

### DIFF
--- a/src/main/kotlin/xyz/bluspring/kilt/loader/remap/KiltRemapper.kt
+++ b/src/main/kotlin/xyz/bluspring/kilt/loader/remap/KiltRemapper.kt
@@ -412,7 +412,13 @@ object KiltRemapper {
             }
 
             withContext(Dispatchers.IO) {
+                val seen = HashSet<String>()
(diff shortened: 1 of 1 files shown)


src/main/kotlin/xyz/bluspring/kilt/loader/remap/KiltRemapper.kt
```diff
@@ -412,7 +412,13 @@ object KiltRemapper {
             }
 
             withContext(Dispatchers.IO) {
+                val seen = HashSet<String>()
                 jar.stream().consumeAsFlow().collect { entry ->
+                    // Ars Nouveau's jar has duplicate entries. wtf?????
+                    if (!seen.add(entry.name)) {
+                        logger.warn("Mod ${mod.id}'s jar file is malformed(duplicate of ${entry.name})!")
```
BluSpring: Needs spacing, also `${mod.displayName} (${mod.id})`

BluSpring (changes requested):

PR #910, opened 2026-07-03
Fix NeoForge custom fuels not working on furnaces

vgskye:


src/main/java/xyz/bluspring/kilt/injects/world/level/block/entity/AbstractFurnaceBlockEntityInject.java
```diff
@@ -27,6 +35,27 @@ public abstract class AbstractFurnaceBlockEntityInject implements AbstractFurnac
         AbstractFurnaceBlockEntityInjection.kilt$tagCookTimes.put(itemTag, burnTime);
     }
 
+    @Unique
+    private RecipeType<?> kilt$recipeType;
```
BluSpring: This should match the actual Neo injected field type and name `RecipeType<? extends AbstractCookingRecipe> recipeType`.

BluSpring (changes requested):

PR #999, opened 2026-09-04
Add some missing injects to ServerGamePacketListenerImpl

### LINKED ISSUE #998: NPC mods that respond to player messages do not work (The Taber, Kerkele)
### Version information

Fabric 1.20.1, Kerkele mod: kerkelemod-0.2-1.20.1.jar, Taber mod: thetaber-1.2.0-forge-1.20.1 EN LANG.jar

### Expected Behavior

Kerkele and Taber both have the function of replying to messages in chat
ie: Taber - Player (in chat): come here. Taber (in reply): What now?
Kerkele - Player (in chat): I love you. Kerkele (in reply): How sweet~

But instead, it's full silence. Every other aspect of the mods work except for them responding to the player in chat. 

### Actual Behavior

Instead of answering, both Taber and Kerkele do not respond at all in chat, or to any messages, even though the correct trigger words are being typed to activate their response
ie Player (in chat): Come here.
Taber doesn't reply. (Come here is a trigger to make him follow you)

Player (in chat): Hello!
Kerkele doesn't respond. (Hello is a trigger word to have him say hello back).

### Reproduction Steps

1: Install Kerkele and/or Taber
2: Summon them properly
3: Try to tell them a command provided by the mod description
4: They will not respond

### Java Provider + Version

check additional information

### Log file

https://mclo.gs/8lTNqXc

### Additional information

I am using a very large modpack, so maybe some weirdness is happening there since there's such a big amount of mods. But, I have a hunch something isn't working with specifically mods from forge that reply to player comments in chat, considering Taber _and_ Kerkele are the only two mods I've had trouble with. All the other forge mods are 100% functional.

(Also my log says Botania and claims do not work together...I don't think that's the issue, they work just fine and came together in this modpack...)

(For my java provider and version, my PC says I have none, I can't find it. I use Prism Launcher, I think it's version 17.0.15, but I don't know the provider)

Edit - sorry I'm new to Github I didn't mean to close it

Acuadragon100:
When checking why the ChatEvent was never fired, I noticed that some other injects for ServerGamePacketListenerImpl were missing. For example, the hand item swap event and positioned player interact events were never fired on the server side. 

For the positioned player interact event I thought it made more sense to use the Porting Lib event, so I removed the client-specific inject for that and let Porting Lib handle both sides.

Closes #998

### DIFF
--- a/src/main/java/xyz/bluspring/kilt/forgeinjects/server/network/ServerGamePacketListenerImplInject.java
+++ b/src/main/java/xyz/bluspring/kilt/forgeinjects/server/network/ServerGamePacketListenerImplInject.java
@@ -5,19 +5,17 @@ import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import com.llamalad7.mixinextras.sugar.Cancellable;
-import com.llamalad7.mixinextras.sugar.Share;
+import com.llamalad7.mixinextras.sugar.Local;
 import com.llamalad7.mixinextras.sugar.ref.LocalRef;
 import net.minecraft.Util;
 import net.minecraft.network.Connection;
 import net.minecraft.network.chat.ChatDecorator;
 import net.minecraft.network.protocol.game.ServerboundCustomPayloadPacket;
-import net.minecraft.network.protocol.game.ServerboundPlayerActionPacket;
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.server.network.ServerGamePacketListenerImpl;
 import net.minecraft.world.InteractionHand;
 import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.common.ForgeHooks;
-import net.minecraftforge.event.entity.living.LivingSwapItemsEvent;
 import net.minecraftforge.network.NetworkHooks;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
@@ -33,43 +31,16 @@ public abstract class ServerGamePacketListenerImplInject {
     @Shadow
     public ServerPlayer player;
 
-    @Inject(
-        method = "handlePlayerAction",
-        at = @At(
-            value = "INVOKE",
-            target = "Lnet/minecraft/server/level/ServerPlayer;setItemInHand(Lnet/minecraft/world/InteractionHand;Lnet/minecraft/world/item/ItemStack;)V",
-            ordinal = 0
-        ),
-        cancellable = true
-    )
-    private void kilt$handleSendPlayerSwapItemsEvent(
-        ServerboundPlayerActionPacket packet, CallbackInfo ci, @Share("event") LocalRef<LivingSwapItemsEvent.Hands> event
-    ) {
-        event.set(ForgeHooks.onLivingSwapHandItems(player));
-        if (event.get().isCanceled()) {
+    @WrapOperation(method = "handlePlayerAction", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/level/ServerPlayer;setItemInHand(Lnet/minecraft/world/InteractionHand;Lnet/minecraft/world/item/ItemStack;)V", ordinal = 0))
+    private void kilt$handleLivingSwapHandItemsEvent(ServerPlayer instance, InteractionHand interactionHand, ItemStack itemStack, Operation<Void> original, @Cancellable CallbackInfo ci, @Local LocalRef<ItemStack> offhandStack) {
+        var event = ForgeHooks.kilt$onLivingSwapHandItems(this.player, offhandStack.get(), itemStack);
+        if (event.isCanceled()) {
             ci.cancel();
+            return;
         }
-    }
 
-    @WrapOperation(
-        method = "handlePlayerAction",
-        at = @At(
-            value = "INVOKE",
-            target = "Lnet/minecraft/server/level/ServerPlayer;setItemInHand(Lnet/minecraft/world/InteractionHand;Lnet/minecraft/world/item/ItemStack;)V"
-        )
-    )
-    private void kilt$updateSwappedItemsFromEvent(
-        ServerPlayer instance, InteractionHand hand, ItemStack itemStack, Operation<Void> original,
-        @Share("event") LocalRef<LivingSwapItemsEvent.Hands> eventRef
-    ) {
-        var event = eventRef.get();
-
-        if (hand == InteractionHand.OFF_HAND && !event.getItemSwappedToOffHand().equals(itemStack)) {
-            itemStack = event.getItemSwappedToOffHand();
-        } else if (hand == InteractionHand.MAIN_HAND && !event.getItemSwappedToMainHand().equals(itemStack)) {
-            itemStack = event.getItemSwappedToMainHand();
-        }
-        original.call(instance, hand, itemStack);
+        original.call(instance, interactionHand, event.getItemSwappedToOffHand());
+        offhandStack.set(event.getItemSwappedToMainHand());
     }
 
     @ModifyExpressionValue(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/xyz/bluspring/kilt/Kilt.kt
```diff
@@ -177,6 +178,14 @@ class Kilt : ModInitializer {
             val chestPiece = entity.getItemBySlot(EquipmentSlot.CHEST)
             chestPiece.canElytraFly(entity) && chestPiece.elytraFlightTick(entity, entity.fallFlyingTicks)
         }
+
+        PlayerInteractionEvents.INTERACT_ENTITY_POSITIONED.register { player, entity, vec3, hand ->
```
BluSpring, line 182: I'm not sure if Porting Lib shares the same injection position that Forge takes. It might be fine but I'm not sure. I'm pretty sure I implemented this in 1.21.1, but again, not sure.

src/main/java/xyz/bluspring/kilt/forgeinjects/server/network/ServerGamePacketListenerImplInject.java
```diff
@@ -16,10 +30,82 @@
...
+    ) {
+        var event = eventRef.get();
+
+        if (hand == InteractionHand.OFF_HAND && !event.getItemSwappedToOffHand().equals(itemStack)) {
+            itemStack = event.getItemSwappedToOffHand();
+        } else if (hand == InteractionHand.MAIN_HAND && !event.getItemSwappedToMainHand().equals(itemStack)) {
+            itemStack = event.getItemSwappedToMainHand();
+        }
+        original.call(instance, hand, itemStack);
+    }
```
BluSpring: Make sure this also handles replacing the local var too, I'm pretty sure Forge does this in their patch. Check Kilt 1.21.1.

src/main/java/xyz/bluspring/kilt/forgeinjects/server/network/ServerGamePacketListenerImplInject.java
```diff
@@ -16,10 +30,82 @@
...
+        cancellable = true
+    )
+    private void kilt$handleSendPlayerSwapItemsEvent(
+        ServerboundPlayerActionPacket packet, CallbackInfo ci, @Share("event") LocalRef<LivingSwapItemsEvent.Hands> event
+    ) {
+        event.set(ForgeHooks.onLivingSwapHandItems(player));
+        if (event.get().isCanceled()) {
+            ci.cancel();
+        }
+    }
```
BluSpring: You can use MixinExtras' `@Cancellable` for this, ref Kilt 1.21.1

src/main/java/xyz/bluspring/kilt/forgeinjects/server/network/ServerGamePacketListenerImplInject.java
```diff
@@ -16,10 +30,82 @@
...
+            target = "Lnet/minecraft/server/MinecraftServer;getChatDecorator()Lnet/minecraft/network/chat/ChatDecorator;"
+        )
+    )
+    private ChatDecorator kilt$combineChatDecorators(ChatDecorator fabricDecorator) {
+        var forgeDecorator = ForgeHooks.getServerChatSubmittedDecorator();
+        return (player, message) -> fabricDecorator.decorate(player, message).thenComposeAsync(
+            fabricMessage -> forgeDecorator.decorate(player, fabricMessage),
+            Util.backgroundExecutor()
+        );
+    }
```
BluSpring, line 59: This slightly differs from Kilt 1.21.1's patch, not sure which one gives a better result.

src/main/java/xyz/bluspring/kilt/forgeinjects/server/network/ServerGamePacketListenerImplInject.java
```diff
@@ -16,10 +30,82 @@
...
+            ci.cancel();
+        }
+    }
+
+    @WrapOperation(
+        method = "handlePlayerAction",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/server/level/ServerPlayer;setItemInHand(Lnet/minecraft/world/InteractionHand;Lnet/minecraft/world/item/ItemStack;)V"
+        )
```
BluSpring: Shouldn't this have an ordinal check?

BluSpring (changes requested):

src/main/java/xyz/bluspring/kilt/forgeinjects/server/network/ServerGamePacketListenerImplInject.java
Acuadragon100: Don't we want this to run for both hands though?

BluSpring: I just realized that what I do for Kilt 1.21.1 is actually a bit different
It runs it only once and changes the local to simplify the mixin
https://github.com/KiltMC/Kilt/blob/version/1.21.1/src/main/java/xyz/bluspring/kilt/injects/server/network/ServerGamePacketListenerImplInject.java#L81-L91

Acuadragon100: Yeah, I basically copied this from LivingEntityInject.
https://github.com/KiltMC/Kilt/blob/version/1.20.1/src/main/java/xyz/bluspring/kilt/forgeinjects/world/entity/LivingEntityInject.java#L285-L304

Acuadragon100: And it looks like 1.21.1 still uses that there.
https://github.com/KiltMC/Kilt/blob/version/1.21.1/src/main/java/xyz/bluspring/kilt/injects/world/entity/LivingEntityInject.java#L641-L660

src/main/java/xyz/bluspring/kilt/forgeinjects/server/network/ServerGamePacketListenerImplInject.java
Acuadragon100: Hmm... Does 1.21.1 handle the case when the returned chat message is null?

Acuadragon100: Actually, I think this currently works the same as the 1.21.1 implementation. In 1.21.1 we currently decorate with the Fabric decorator first, then the result of that with the Forge one. https://github.com/KiltMC/Kilt/blob/version/1.21.1/src/main/java/xyz/bluspring/kilt/injects/server/network/ServerGamePacketListenerImplInject.java#L98

It's just that here the ChatDecorator expects the result as a CompletableFuture and the code runs asynchronously.

src/main/kotlin/xyz/bluspring/kilt/Kilt.kt
Acuadragon100: From what I can see, it is basically identical on the server side.
On the client side, Forge seems to inject before the spectator mode check while Porting Lib injects after it. But Forge also adds their own separate spectator mode check before that so I don't think it matters.

Acuadragon100: Forge clientside:
https://github.com/KiltMC/NeoForge/blob/kilt/1.20.1/patches/minecraft/net/minecraft/client/multiplayer/MultiPlayerGameMode.java.patch#L165-L171

Porting Lib clientside:
https://github.com/Fabricators-of-Create/Porting-Lib/blob/1.20.1/modules/entity/src/main/java/io/github/fabricators_of_create/porting_lib/entity/mixin/client/MultiPlayerGameModeMixin.java#L55-L68

Forge serverside:
https://github.com/KiltMC/NeoForge/blob/kilt/1.20.1/patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch#L87-L92

Porting Lib serverside:
https://github.com/Fabricators-of-Create/Porting-Lib/blob/1.20.1/modules/entity/src/main/java/io/github/fabricators_of_create/porting_lib/entity/mixin/common/ServerGamePacketListenerImplInteractHandlerMixin.java#L19-L32

BluSpring: Oh that's actually pretty much identical

PR #1009, opened 2026-09-19
Fix annotations getting duplicated when applying coremods.

Acuadragon100:
Not sure if this affects production at all, but it caused a crash in dev (when combining Cobblemon and Twilight Forest) and it bugged me.


Acuadragon100:
Actually, yes the crash happens in production as well.

